PR #36, opened 2019-03-25
Derive targets

jacksonsj:
Code to derive VMMC_CIRC Age/Sex from Site Tool.

Also fixes issue with site tool to DATIM mapping

Also fixes issue with defunct disaggs flagging OVC_HIVSTAT based on Data Pack config instead of Site Tool


R/parseSiteTool.R
```diff
@@ -350,7 +350,8 @@ unPackSiteToolSheet <- function(d) {
     # Drop zeros, NAs, dashes, and space-only entries
     tidyr::drop_na(value) %>%
     dplyr::filter(
-      !is.na(suppressWarnings(as.numeric(value)))) %>%
+      !is.na(suppressWarnings(as.numeric(value)))
+      & value != 0) %>%
```
jason-p-pickering: I am pretty sure we do not want to drop all zeros. Zeros for dedupe are actually significant for import. We can drop all non-dedupe zeros, but need to account for that particular case. I reorganized slightly here.

R/parseSiteTool.R
```diff
@@ -451,3 +453,53 @@ unPackSiteToolSheet <- function(d) {
...
+#' 
+#' @return Dataframe with added, derived targets.
+#' 
+deriveTargets <- function(data, type) {
+  derived <- data %>%
+    dplyr::filter(
+      stringr::str_detect(
+        indicatorCode, 
+        paste0(
+          "VMMC_CIRC\\.N\\.Age/Sex/HIVStatus\\.20T"
```
jason-p-pickering: Not sure I understand what this does, but maybe it works?

PR #37, opened 2019-03-29
Compare sitetool to datapack

jacksonsj:
See comparePacks.R.

ALSO, the na.omit added in parseSiteTool is dropping entire sheets where data is na in one column, but not in all others. Had to remove this to allow parsing to occur.

Switched to this in the middle of work moving data pack writing into the package, so including that here as well. The new pieces included are largely self-contained.


R/comparePacks.R
```diff
@@ -0,0 +1,148 @@
...
+      dplyr::mutate(diff = (value.sitetool - value.datapack)/value.datapack) %>%
+      dplyr::arrange(dplyr::desc(abs(diff)))
+     
+  # Write to Excel Document
+    wb <- openxlsx::createWorkbook()
+    
+    sheetNames <- c("Data Pack Data", "Site Tool Data", "Comparison", "Diffs",
+                    "Category Summary", "Indicator Summary")
+    
+    invisible(sapply(sheetNames, function(x) openxlsx::addWorksheet(wb, sheetName = x)))
```
jason-p-pickering: Seems a bit overly complex to me. Why not create a list object of all of the data frames and just write them?

R/datapackr.R
```diff
@@ -23,17 +23,21 @@ if (getRversion() >= "2.15.1")
   utils::globalVariables(
     c(
       ".",
+      ":=",
```
jason-p-pickering: Hmm...what is the purpose of this?

R/packDataPack.R
```diff
@@ -0,0 +1,83 @@
...
+                     datapack_uid = d$info$datapack_uid,
+                     type = "Data Pack")
+  
+  # Get PSNU List####
+  d$data$PSNUs <- getPSNUs(country_uids = d$info$country_uids,
+                           include_mil = TRUE)
+  
+  # Write Main Sheets ####
+  sheets <- readxl::excel_sheets(d$keychain$template_path)
+  sheets_to_loop <- sheets[which(!stringr::str_detect(sheets, "Home|Quotes|Summary|Spectrum|SNU x IM|Visualizations|Validations"))]
```
jason-p-pickering: Anyway to put this into a schema/configuration file instead of hardcoding it?

R/parseSiteTool.R
```diff
@@ -303,17 +303,17 @@ unPackSiteToolSheet <- function(d) {
...
-      na.omit(.) %>%
+      stats::na.omit(.) %>%
       NROW(.) == 0
   }
   
-  if ( is_empty_data_frame(d$data$extract) ) {
-    d$data$extract <-NULL
-    return(d)
-  }
+  # if ( is_empty_data_frame(d$data$extract) ) {
```
jason-p-pickering: OK, i just added this code to deal with situations where the sheet was completely empty, which caused other problems. Seems it is not working as intended?

R/utilities.R
```diff
@@ -280,6 +280,17 @@ exportPackr <- function(data, output_path, type, datapack_name) {
 }
 
 
+#' @export
```
jason-p-pickering: I think its fine to export this, but if you are interactively coding, you can use the `:::` triple colon operator to call functions which are not exported.

R/utilities.R
```diff
@@ -292,18 +303,23 @@ interactive_print <- function(x) if (interactive()) { print(x) }
...
+                     include_mil = FALSE) {
   
   countries <- datapackr::api_call("organisationUnits") %>%
     datapackr::api_filter("organisationUnitGroups.id:eq:cNzfcPWEGSH") %>%
     datapackr::api_fields("id,name,level") %>%
     datapackr::api_get()
   
   PSNUs <- datapackr::api_call("organisationUnits") %>%
-    datapackr::api_filter("organisationUnitGroups.id:in:[AVy8gJXym2D,nwQbMeALRjL]") %>%
+    datapackr::api_filter(paste0("organisationUnitGroups.id:in:[AVy8gJXym2D",
```
jason-p-pickering: I think it would be tidier not to have these UIDs hardcoded, or at least as a local variable.

DESCRIPTION
```diff
@@ -38,7 +38,9 @@ Imports:
     openxlsx,
     lazyeval,
     stats,
-    lubridate
+    lubridate,
+    data.table,
```
jason-p-pickering: Do we really need this dependency?

PR #44, opened 2019-04-30
Align datapack sitetool code

jacksonsj:
@jason-p-pickering & @sam-bao - Could you both review this code and send feedback?

The purpose of these changes was to better align code used to process Data Packs and Site Tools, which had been largely duplicative and clunky.

Because of the many downstream implications of some of these changes, I'd like to ask for both of you to review to identify ways these changes may impact your code.

**Major changes include:**
- Tagging country UIDs based on the processed data itself, rather than based on tool home page. This allows for all variations of country pairings to occur in the Site Tool or Data Pack. But, this may have downstream implications on validation apps.
- Functionalizing code processes and forking to allow usage in both Data Pack and Site Tool processing with variation based on need.
- Removing some warnings that had been issued before in integrity checks that have confused users and are not actually informative (e.g., column or tab order issues).
- Functionalizing process used to test and open a file.
- Grouping all tool processing, whether Data Pack or Site Tool, under `unPackTool`.
- Standardizing naming of `indicator_code` (instead of `indicatorCode`)


R/defunctDisaggs.R
```diff
@@ -0,0 +1,77 @@
...
+#'    been used.
+#'
+#' @param d datapackr list object containing at least d$data$extract of data
+#'     from a single sheet in submission file.
+#'     
+#' @return A datapackr list object, \code{d}, storing a dataframe of defunct
+#'    disaggs used in a specific sheet.
+defunctDisaggs <- function(d) {
+  
+  defunct <- d$data$extract %>%
```
jason-p-pickering: Maybe it would make sense if possible to have this as a configuration file? We know there are likely going to be defunct disaggs next year, so a more flexible combination of indicator code and disagg might work?

R/parseDataPack.R
```diff
@@ -379,116 +127,140 @@ unPackSheet <- function(d) {
...
+  
   # TEST for Negative values
   has_negative_numbers <- d$data$extract$value < 0
   
   if (any(has_negative_numbers)) {
-    negCols <- d$data$extract %>%
+    neg_cols <- d$data$extract %>%
       dplyr::filter(value < 0) %>%
-      dplyr::pull(indicatorCode) %>%
+      dplyr::pull(indicator_code) %>%
```
sam-bao: change to indicator code will require very minor change to datapack commons code (https://github.com/pepfar-datim/data-pack-commons/search?q=indicatorCode&unscoped_q=indicatorCode) I added an issue to that repo (https://github.com/pepfar-datim/data-pack-commons/issues/22)

R/checkOUinfo.R
```diff
@@ -0,0 +1,93 @@
...
+#' @param d datapackr list object containing at least d$keychain$submission_path.
+#' @return A datapackr list object, \code{d}, storing a unique UID and Name for
+#'    the PEPFAR Operating Unit related to the submitted Data Pack or Site Tool.
+checkOUinfo <- function(d) {
+  # Get OU name and uid
+  d$info$datapack_uid <-
+    names(readxl::read_excel(
+      d$keychain$submission_path,
+      sheet = "Home",
+      range = "B25"
```
jason-p-pickering: Is there anyway we could get this out as a hardcode and into a schema/configuration?

sam-bao:
I unpacked a Lesotho data pack and completed a diff against the original unpacked data pack 
for data$distributedMER and the only difference found was the name change from indicatorCode to indicator_code.

I also checked datim$PSNUxIM and datim$SUBNAT_IMPATT and found no differences.

sam-bao:
@jacksonsj I agree with Jason's comments about pulling out some of the hard coded bits that are tightly coupled to COP19 and placing in a config file. 

Additionally we have already discussed the need to improve our naming across the board. Lines like this are a pretty good example of our inconsistency. 
 
```dplyr::select(site_uid,mech_code,indicator_code,Type,Age,Sex,KeyPop,value) ```

This should probably be
```dplyr::select(site_uid, mechanism_code, indicator_code, support_type , age, sex, key_population, value) ```

That said I think we can do a general renaming once the main refactoring is done.

File: R/unPackSiteTool.R [hunk 4]
```diff
@@ -1,108 +1,51 @@
...
-  d$info$has_error <- FALSE
+  site_uids <- d$data$targets %>%
+    dplyr::pull(site_uid) %>%
+    unique() %>%
+    paste0(collapse=",")
   
-  # Check OU name and UID match up
-  interactive_print("Checking the OU name and UID on HOME tab...")
-  d <- checkSiteToolOUinfo(d)
+  countries <-
```
sam-bao: I have a function for metadata calls in datapack commons (https://github.com/pepfar-datim/data-pack-commons/blob/f1bedc5bc1a7c25e2a1993920d0c934813e68242/R/dhis2api.R#L278)

I wanted to make a small change to this function to move base_url to the end with a default of getOption("baseurl")

This would make the equivalent call something like:

```
datapackcommons::getMetadata(end_point = "organisationUnits", 
                                                         filters = c("organisationUnitGroups.name:eq:Country",
                                                                           paste0("children.id:in:[",site_uids,"]")),
                                                         fields = "id,name")
``` 
We should coordinate how we want to make common api calls like this. I bet Jason has his own version somewhere as well.

R/unPackSiteTool.R
```diff
@@ -1,108 +1,51 @@
...
+      d$keychain$submission_path,
+      sheet = "Home",
+      range = "B20") %>%
+    names()
   
-  # Start running log of all warning and information messages
-  d$info$warningMsg <- NULL
-  d$info$has_error <- FALSE
+  site_uids <- d$data$targets %>%
+    dplyr::pull(site_uid) %>%
```
sam-bao: Any chance this will result in too many sites to make the api call - there is some maximum length to the api call.

R/unPackData.R
```diff
@@ -36,6 +36,18 @@
 
 unPackDataPack <- function(d) {
 
+  # Determine country_uids
+    if (is.na(d$info$country_uids)) {
+      d$info$country_uids <-
+        readxl::read_excel(
+          path = d$keychain$submission_path,
+          sheet = "Home",
+          range = "B23") %>%
```
sam-bao: @jacksonsj I think currently ou uid is in B25, has this changed?

sam-bao:
@jacksonsj for these most recent changes. I was able to unpack Lesotho data pack without issue, and confirmed the distributedMer output, SNUxIM output, and datim$PSNUxIM matched the original unpacking I did for Lesotho.

sam-bao:
@jacksonsj is this still a valid pull request/branch? As this makes a lot of changes I don't think we can merge it into master at this moment - I think it represent post cop 19 refactoring. Should we close the pull request without merging?

sam-bao:
@jacksonsj Can we close this pull request and delete the branch?

PR #48, opened 2019-05-15
Regional fix sam

sam-bao:
This pull requests allows getSiteList to function when provided a region uid.

I was unable to run a compare on the Central Asia site tool and data pack. The Central Asia data pack contained 3 countries in the asia region. I was trying to run the compare while having the Asia region uid on the home page in cell B25 for both the datapack and site tool, but it was not working. I traced the problem to getSiteLists not accepting a region.

I also manually changed a number of military org unit uids from TBD... to real uids in the datapack. I am unsure if this was required in the end as it was one of the first things I tried to make the compare work.

I was just trying to get the compare working correctly, I can clean this up a bit if we want to merge it.


sam-bao:
@jason-p-pickering you previously gave a green light for this pull request on Slack. I would like to pull this into the regional fix branch as a precursor to pulling the regional-fix branch into datapackr master.

PR #54, opened 2019-09-02
GetDataValueSets function initial committ

sam-bao:


sam-bao:
@jason-p-pickering 

It took a bit of trial and error, but I now have the dependencies specified such that the travis build is successful on this PR for the DataValueSets function.

PR #71, opened 2019-12-11
Update Generate19From20TargetData.R

sam-bao:
Put mismatched records in another object for now, will need to be insetigated case by case.

Modify code to use results of getSiteList call and SiteToDatim table to go from data value set to data packr `d` object.

## Affected Process Elements
- [ ] Data Pack generation
- [ ] Data Pack model
- [ ] Data Pack self-service app
- [ ] Data Pack processing
- [X] Site Tool generation (from DATIM - for OPUs)
- [ ] Site Tool processing
- [ ] Site Tool import

## Types of changes

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes)
- [ ] Build-related changes

- [ ] Yes
- [x] No

### DIFF
--- a/data-raw/Generate19From20TargetData.R
+++ b/data-raw/Generate19From20TargetData.R
@@ -87,6 +87,7 @@ getCOP19DataFromAPI<-function(country_uid) {
     "dataSet", "HiJieecLXxN",
     "period",  "2019Oct",
     "children", "true",
+    "categoryOptionComboIdScheme", "code",
     "includeDeleted", "false",
     "orgUnit", country_uid) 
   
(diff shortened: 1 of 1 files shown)


data-raw/Generate19From20TargetData.R
```diff
@@ -87,6 +87,7 @@ getCOP19DataFromAPI<-function(country_uid) {
     "dataSet", "HiJieecLXxN",
     "period",  "2019Oct",
     "children", "true",
+    "categoryOptionComboIdScheme", "code",
```
sam-bao: I added this to get the mechanism code instead of category option uid.

data-raw/Generate19From20TargetData.R
```diff
@@ -102,12 +103,57 @@ getCOP19DataFromAPI<-function(country_uid) {
 
 datimvalidation::loadSecrets(secrets)
 
+# get a table to map the DATIM columns data site tool columns 
+# handling the special cases of HTS_SELF.N.HIVSelfTest.20T.Unassisted and 
+# 
+ site_to_datim <- dplyr::filter(datapackr::SiteToDATIM,
```
sam-bao: This is to take care of some edge cases. When we originally imported data from the site tools some data that is disaggregated in the site tool was aggregated when placed in Datim. We cannot disaggregate these fro the OPU site tools so instead we are selecting one line item for the aggregated data to go to. When we join site_to_datim to the data value set we should only get one match per record in dataSets

data-raw/Generate19From20TargetData.R
```diff
@@ -102,12 +103,57 @@ getCOP19DataFromAPI<-function(country_uid) {
...
+                         (indicator_code == "HTS_SELF.N.HIVSelfTest.20T.Unassisted" &
+                            valid_ages == "50+" & valid_sexes == "Male")
+                       ) %>% 
+   dplyr::filter(indicator_code != "TB_ART.N.Age/Sex/NewExistingART/HIVStatus.20T.Already")
+   
+
 # Cache the PSNUs for this country
-psnus = datapackr::getPSNUs(parent_uid, TRUE)
+# psnus = datapackr::getPSNUs(parent_uid, TRUE)
+sites  <-  datapackr::getSiteList(parent_uid)
```
sam-bao: this getSiteList function brings back all sites already joined to its respective PSNU. I think it is best to reuse this instead of looping over all rows in dataSets to find an org unit's respective psnu. We can easily join dataSets with this sites object.

getSiteLists is very slow and could probably use a refactoring, but it is fit for purpose in the sense of OPUs.

data-raw/Generate19From20TargetData.R
```diff
@@ -102,12 +103,57 @@ getCOP19DataFromAPI<-function(country_uid) {
...
 # Store the country OU name
 parent_name <- getOUNameFromUID(parent_uid)
 # Get the main dataset for looping through
 dataSets <- getCOP19DataFromAPI(parent_uid)
+
+#join with site to datim table to back out columns for site tool 
+dataSets  <-  dplyr::left_join(dataSets, site_to_datim,
```
sam-bao: Here I join dataSets with site_to_datim which gives un the disaggregation information we will need to build `d`

data-raw/Generate19From20TargetData.R
```diff
@@ -102,12 +103,57 @@ getCOP19DataFromAPI<-function(country_uid) {
...
 dataSets <- getCOP19DataFromAPI(parent_uid)
+
+#join with site to datim table to back out columns for site tool 
+dataSets  <-  dplyr::left_join(dataSets, site_to_datim, 
+                 by = c("data_element" = "dataelementuid",
+                        "category_option_combo" = "categoryoptioncombouid"))
+
+# join with sites table to get PSNU details
+
+dataSets  <-  dplyr::left_join(dataSets, sites,
```
sam-bao: Here I join with sites to get the psnu information.

data-raw/Generate19From20TargetData.R
```diff
@@ -102,12 +103,57 @@ getCOP19DataFromAPI<-function(country_uid) {
...
+# join with sites table to get PSNU details
+
+dataSets  <-  dplyr::left_join(dataSets, sites, 
+                                by = c("org_unit" = "id"))
+
+# store these for investigation - Why is there FY20 target data in DATIM 
+# that we cannot map to the site tool? Usually this is because the data collection form included
+# entry points that were not on the data pack and someone manually eneterd the target
+
+incomplete_records = dplyr::filter(dataSets, is.na(indicator_code) | is.na(psnu))
```
sam-bao: These are most likely bad data due to mismatches between the site tools and the targets data collection form. Ocasionally countries were authorized to manually change target data in DATIM and the sometimes fill in these cells which should be empty on the data collection form.

data-raw/Generate19From20TargetData.R
```diff
@@ -102,12 +103,57 @@ getCOP19DataFromAPI<-function(country_uid) {
...
+
+dataSets = dplyr::filter(dataSets, !is.na(indicator_code) & !is.na(psnu)) %>% 
+  dplyr::select(PSNU = psnu,
```
sam-bao: After the joins I can simply select the columns i need for `d`

sam-bao:
@Awnage I pulled your code into the datapackr package as a branch, and this pull request includes some changes I propose for the code to reuse information/functions already in the package.

Some additional comments. We use snake_case for variables and objects other than functions and camelCase for function names.

It looks like you are going down the path of looping over rows of a data set, but it is best practice in R to use functional programming in these cases (https://towardsdatascience.com/cleaner-r-code-with-functional-programming-adc37931ef7a, https://r4ds.had.co.nz/iteration.html, https://purrr.tidyverse.org/).

Please have a look, and let me know if you have any questions

PR #87, opened 2020-01-16
Fix decimal value check

### LINKED ISSUE #64: False Positives for decimal values in unPackSheet
This line of code is not reliably detecting non-integers.

https://github.com/pepfar-datim/datapackr/blob/9b34ce5775f503a1a6250a148b17a6a2da4f232e/R/unPackSheet.R#L133

As an example for this data pack: https://www.pepfar.net/Project-Pages/collab-38/Shared%20Documents/Data%20Pack%202019%20Staging%20Area/Support%20Files/datapacks/71_DataPack_Uganda_20190124160453_03082019.xlsx,

non decimals are flagged on the PMTCT_STAT_ART tab in the `PMTCT_STAT.D.Age/Sex.20T` column. However looking at the excel version of the data pack does not reveal any non-integer numbers. There appears to be some floating point error introduced when  readxl::read_excel initially reads in the sheet.

image

sam-bao:
fixes #64

Reading sheets from excel is bringing along some small floating point error. This pull request adds rounding of values to 10 decimal places in an attempt to eliminate this small rounding error without compromising a real test for decimal vs integer.


sam-bao:
@jason-p-pickering @jacksonsj I am now running into this same issue with last years site tool validation app.

Extracting the data as we do in the code and filtering on military
```
data <-
  readxl::read_excel(
    path = file.choose(), # i am using the mzambique OPU site tool returnd 
    #https://datim.zendesk.com/attachments/token/fg611XC1S1B5DmrCtbHcX0hJR/?name=COP19_OPU_Site+Tool_Mozambique_20200107153120_MT1.8.19+JL.xlsx
    sheet = "VMMC",
    range = readxl::cell_limits(c(5, 1), c(NA, NA))
  )

print(dplyr::filter(data, stringr::str_detect(Site, "Military")))
```

Results in a non-integer value in what was initially an integer in the site tool.

image


image

Extra decimals added to the excel to be sure there isn't a hidden decimal in the origin file.

sam-bao:
Some more details related to this issue: https://github.com/tidyverse/readxl/issues/37

and we do appear to read in the data as text specifically now: https://github.com/pepfar-datim/datapackr/blob/643d9190597c31d6726002a5e59b09eae8702c5f/R/unPackSheet.R#L235

PR #89, opened 2020-01-16
Tests storage fix

### LINKED ISSUE #61: Bug if >1 column contains decimal values
Receiving this error:

```
Error in d[["tests"]][["decimal_cols"]][[as.character(sheet)]] <- decimal_cols : 
  more elements supplied than there are to replace
```

trying to parse this datapack:

https://www.pepfar.net/Project-Pages/collab-38/Shared%20Documents/Data%20Pack%202019%20Staging%20Area/Support%20Files/datapacks/DataPack_Namibia_20190314_1200.xlsx

The code I used to parse the data pack:
```
country_uids <- c("FFVkaV9Zk1S")

submission_path <- "###"
## Note that submission_path is optional in this setup. If not supplied, a console window will pop up to allow you to pick the file.

d <- datapackr::unPackTool(submission_path = submission_path,
                tool = "Data Pack",
                country_uids = country_uids)
```

Problem is related to this line of code:
https://github.com/pepfar-datim/datapackr/blob/e6553e729890adf2f57374991da9a35b1df421ff/R/unPackSheet.R#L140

It is not obvious to me what should go in to `d[["tests"]][["decimal_cols"]][[as.character(sheet)]] ` but it seems like this works:

```    
d[["tests"]][["decimal_cols"]][[as.character(sheet)]] <- list(decimal_cols)
```

This same issue may be repeated in other pieces of code such as
```
 d[["tests"]][["non_numeric"]][[as.character(sheet)]] <- non_numeric
```

sam-bao:
If the first sheet processed with an error of a given type had only one error then the d[["tests"]][["#type#"]] object is initialized as a character vector instead of a list. If a subsequent sheet has multiple columns with errors of this type we get an error when we try to add a vector for d[["tests"]][["#type#"]][[#sheet#]] to the character vector object. E.g. `Error in d[["tests"]][["non_numeric"]][[as.character(sheet)]] <- non_numeric :  more elements supplied than there are to replace`

Initializing each test type object with an empty character vector will force each d[["tests"]][["#type#"]] object to be of type list even if the first error of a given type is attributable to only one column.


sam-bao:
@jacksonsj I am still having problems parsing a data pack that are resolved with this fix. Please have a look and if the fix looks good merge it.

The latest issue is when I try to parse this data pack: https://www.pepfar.net/ou/rwanda/HQ%20Collaboration/COP%202020%20%E2%80%93%20FY%202021/Guidance%2C%20Tools%2C%20and%20Resources/Data%20Pack_Rwanda_20200120225346.xlsx

Here is the result:
```
datapackr::unPackTool()
[1] "Please choose a file."
[1] "Checking for any missing tabs..."                                                    
[1] "Epi Cascade I"
[1] "Epi PMTCT"                                                                           
[1] "Prioritization"                                                                      
[1] "TX"                                                                                  
[1] "HTS"                                                                                 
Error in d[["tests"]][["non_numeric"]][[as.character(sheet)]] <- non_numeric :            
  more elements supplied than there are to replace
```

sam-bao:
@jason-p-pickering Are you able to review this and let me know if I am free to squash and merge?

sam-bao:
After discussion with Jason, I will merge this now, and open a ticket to revisit these lines of code later. There should be a better way to ensure we have a list object here.

PR #90, opened 2020-01-16
Resolve some issues resulting from Tidyr upgrade

sam-bao:


jason-p-pickering:
Still need to see what the renaming from name1->categoryname will break. That is pending, so for now, do not merge this.

sam-bao:
OBE

PR #147, opened 2020-03-25
gh699 and 700 and 709

sam-bao:
Includes FY21/COP20 specific code for getting COP20 data from DATIM after it has been imported.

This functionality was required for the following purposes:

1) Global Fund data transfer
2) Deletion file in the case a countries data pack needs to be reloaded
3) data pack vs datim comparison.

resolves 
https://github.com/pepfar-datim/COP-Target-Setting/issues/699
https://github.com/pepfar-datim/COP-Target-Setting/issues/700
https://github.com/pepfar-datim/COP-Target-Setting/issues/709


sam-bao:
The functionality in https://github.com/pepfar-datim/COP-Target-Setting/issues/699 is required now but the approach for https://github.com/pepfar-datim/COP-Target-Setting/issues/700 is being reconsidered based on https://github.com/pepfar-datim/COP-Target-Setting/issues/734.

I will pull the changes specific to #700 into a separate branch.

PR #167, opened 2020-07-30
Remove select OUs function

sam-bao:
@jacksonsj 

This function is not used anywhere in datapackr, and as it is interactive we can safely assume it is not used in any shiny apps. I can't really see a use for this function going forward.

This change is related to https://github.com/pepfar-datim/datapackr/pull/166/files and https://github.com/pepfar-datim/COP-Target-Setting/issues/844
.


sam-bao:
@jacksonsj reminding you about this PR.

sam-bao:
OBE

PR #207, opened 2020-12-21
add httr::timeout(180) to all httr::GET

sam-bao:
resolves https://github.com/pepfar-datim/Data-Pack/issues/1087


jason-p-pickering (commented):
I checked quickly with the OPU app (where this was happening) and it seems to be fine. I will merge it.

PR #216, opened 2021-01-13
Add method to bootstrap COP year and tool type

### LINKED ISSUE #215: Try and get type of tool and COP year if not specified
The current OPU and Datapack app share basically the same code and functionality. Keeping both of these apps maintained will be laborious and duplicative.  With one app, we should be able to perform the necessary validations on both OPU DataPacks and normal DataPacks, since the vast majority of the code is essentially the same. 

We should be able to pretty easily determine what type of tool we are working with, and from there, decide what to do with it in the app. Ideally, we could write the specific type of tool "Data Pack", "OPU Data Pack", etc into a specific range of cells on the Home tab, but this is currently available in cell  B10, like "COP21 Data Pack" or "COP20 OPU Data Pack". Once we have this information in the app, we can proceed with the specific processing which each tool requires. 

Command line users or apps would still be able to specify this information for specific use cases, but if left blank (NULL) we would try and obtain this information from the home tab.

jason-p-pickering:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

Attempt to get the COP year and tool type if it is not specified. 

## Affected Process Elements
- [ ] Data Pack generation
- [ ] Data Pack model
- [X] Data Pack self-service app
- [ ] Data Pack processing
- [ ] Site Tool generation (from DATIM - for OPUs)
- [ ] Site Tool processing
- [ ] Site Tool import

## Types of changes

- [ ] Bugfix
- [X] Feature
- [ ] Refactoring (no functional changes)
- [ ] Build-related changes

- [ ] Yes
- [X] No

## Related Issues

- Resolves #215


sam-bao:
@jason-p-pickering I have made a few changes to this PR. I don't think Scott needs to review this any further, so unless ou have any questions or comments on my additonal changes I think we can merge this.

My changes where mostly related to removing the default function argument of `tool = "Data Pack"` in several places in order to ensure we actually make use of the bootstrapped tool details and also ensuring that d$info$cop_year continues to be numeric even after we scrap it (as char) from the datapack.

jason-p-pickering:
Thanks for the review and changes @sam-bao . I have retested with the app, and it seems to work as intended.  Merging and closing.

PR #219, opened 2021-01-18
Update packSNUxIM.R

jason-p-pickering:
There is a lingering local variable reference in PackPSNUxIM, which should instead reference `d$info$cop_year`. 

This prevents processing of the PSNUxIM tab in the app. 

FYI @jacksonsj 

@sam-bao if you are OK with this change, please merge, as its preventing further work on the app.

### DIFF
--- a/R/packSNUxIM.R
+++ b/R/packSNUxIM.R
@@ -78,7 +78,7 @@ packSNUxIM <- function(d) {
                                        -psnuid, -sheet_name))
           
       # Allow DataPackTarget formula to lookup KP_MAT correctly ####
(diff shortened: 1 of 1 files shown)


sam-bao:
@jason-p-pickering the SNUxIM code is under active development in branch 4.1, the code you changed 1) doesn't seem to exist in 4.1 and 2) creates a bizarrely large conflict with branch 4.1.

I do not think we should merge this into master. Can you check and see if branch 4.1 advances or pauses you current dev task.

sam-bao:
As this seems to be fixed in 4.1 closing without merging into master

PR #223, opened 2021-01-25
Create cop20 map data pack datim d es co cs

sam-bao:
Create a `cop20_map_DataPack_DATIM_DEs_COCs` and modify code to use it where applicable. This is required because in release 4.1 `map_DataPack_DATIM_DEs_COCs`is updated and becomes invalid for processing OPUs.

https://github.com/pepfar-datim/Data-Pack/issues/1099

This should be merged into master and then master should be pulled into release 4.1.

I have confirmed that the output is identical between master and this branch when generating and parsing a COP20 OPU and COP21 datapack.


R/create_play_spectrum_output.R
```diff
@@ -18,6 +18,15 @@
 create_play_spectrum_output <- function(country_uids,
                                         cop_year,
                                         output_folder = NULL) {
+  
+  if (cop_year == 2021) {
+    map_DataPack_DATIM_DEs_COCs_local <- datapackr::map_DataPack_DATIM_DEs_COCs
```
jason-p-pickering, line 23: maybe better to create a function here? `mapDATIMDesCocs(cop_year=2021)`

sam-bao: Yep, I will make that change.

sam-bao: @jason-p-pickering I added a ticket to address your comment. I would like to go ahead and merge this PR into master and then 4.1 and address this refactoring in either 4.1 or as a fast follow to 4.1. Getting 4.1 is my priority and the changes in this pull request, while redundant, are functional.

https://github.com/pepfar-datim/data-pack/issues/1102

jason-p-pickering: :+1:

jacksonsj (approved):

PR #236, opened 2021-02-08
unpackSchema exclude FY column when cop_year==2020

sam-bao:
Unable to pack a COP20 OPU due to presence of FY in schema. This excludes FY if cop_year == "2020"


jacksonsj (approved):
Approved.

PR #242, opened 2021-02-17
Gh1129

maxwellchandler:


maxwellchandler:
@sam. Yes Ive done that. Note that all the tests pass when running covr::report() or using testthat::test_dir("tests/testthat/") but one tests fails when using devtest:test() and im not sure why. I prefixed this test with a 1 so it runs first.

maxwellchandler:
There is just one file called helper-variables.R in the test directory that sets the paths

PR #246, opened 2021-02-24
Add unit tests for round_trunc

jason-p-pickering:
Adds unit tests from the datapack-importer package for the round_trunc function.


jacksonsj:
@jason-p-pickering - Let's wait on this one until after PR #257 is in `dev`. It may also be a good idea to compare this against PR #248 , which introduces seismic changes in unit testing, to make sure the two approaches align.

jacksonsj (approved):

jacksonsj:
@jason-p-pickering - PR #257  is in `dev` and I approve this PR to go in as well, but would like @maxwellchandler to review as well given the overlap with his work on unit tests right now

PR #249, opened 2021-03-02
Update update_cached_de_coc_co_map.R

sam-bao:
Minor bug fix in parameters for pivot_wider


sam-bao:
This was the result of me using an older version of tidyr, so I am closing without merging.

PR #256, opened 2021-03-09
remove getCOPDatasetUIDS

sam-bao:
@jacksonsj This function getCOPDatasetUIDS was replaced by getDatasetUids a few releases ago. This PR removes getCOPDatasetUIDS. All the code that used to use it was updated here: https://github.com/pepfar-datim/datapackr/pull/251.


jacksonsj:
Yeah looks like `getCOPDatasetUIDs` was at one point used within `getCOPDataFromDATIM`, but has been replaced with `getDatasetUids`

PR #259, opened 2021-03-15
Refactor adorn mechanisms

jason-p-pickering:
https://jira.pepfar.net/browse/DP-186 for details.


jacksonsj (commented):
@jason-p-pickering - Can we discuss over phone so I can understand the issues we were seeing before with server-side filtering, how this fits into the S3 plan, and how this will get triggered by the Shiny app?

jacksonsj (approved):

PR #260, opened 2021-03-15
Compare branch

maxwellchandler:
this is the separate comparison branch. Note that due to the same reasons that getAnalytics does not work with the test files in dev, the current script will fail. However the base is there.


sam-bao:
This is stale OBE, I am keeping the branch in place for now, but am closing the PR.

PR #264, opened 2021-03-24
Update datapackVsDatim.R to include previous year subnat and impatt

sam-bao:
@jacksonsj This updates the datapack vs datim comparison to also include the FY21 subnat and impatt in the comparison


jacksonsj (approved):

jacksonsj:
@sam-bao - I approve. Yes, this makes sense. Good to go into `dev`

PR #269, opened 2021-04-09
Create COP21_datapacks_countries object

sam-bao:
Add a package data object that provides the country groupings for each COP21 datapack


jason-p-pickering:
Look good. Thanks for this. Merging.

PR #271, opened 2021-04-18
unPackSchema for COP20 exclude the new period column

sam-bao:
period was added as part of PR267, but isn't applicable for COP20 OPU datapacks

fix for https://jira.pepfar.net/browse/DP-223


jacksonsj (approved):
Approved

PR #273, opened 2021-04-20
Write all formulas on right side of PSNUxIM tab

jacksonsj:
- Automates writing all formulas to right side of PSNUxIM tab.
- Does not fully address all performance implications of this automation.

### DIFF
--- a/R/packSNUxIM.R
+++ b/R/packSNUxIM.R
@@ -37,6 +37,12 @@ packSNUxIM <- function(d,
   }
 
   # Prepare SNU x IM model dataset ####
+  if (d$info$has_psnuxim & d$info$missing_psnuxim_combos) {
+    targets_data <- d$data$missingCombos
+  } else {
+    targets_data <- d$data$MER
+  }
+  
     #TODO: Consider preparing this ahead of time for all OUs
   snuxim_model_data <- readRDS(d$keychain$snuxim_model_data_path) %>%
     prepare_model_data.PSNUxIM(model_data = .,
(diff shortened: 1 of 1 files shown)


jacksonsj:
@jason-p-pickering - Could you review this to include in `dev`?

jason-p-pickering:
@jacksonsj I tested this branch with the app, and did not see any difference in the formulas in the columns to the right. Could you confirm that this patch does what its supposed to?

I think we share concerns about the amount of time which this will add to the processing of the PSNUxIM tabs, so having some indication of how much longer it will take would be good as well. On my local machine with the app, the Malawi datapack took about 1 minute 40 seconds to generate. Could you include the differences in processing time as well, once you confirm things are working as intended?

jacksonsj:
@jason-p-pickering - I think we discussed this on our standup last week, but this only automates dragging formulas down for any rows actively being edited by the app. So if a file already has rows written in, and the app is only adding to the bottom, this script change only adds formulas for those rows added to the bottom — not the others previously there.

Any hesitations on me pushing this to `dev`?

jason-p-pickering:
No. I agree that this address the situation which you outline.  

Just thinking out loud however, could we just add an integrity check here to check for the presence of the formulas? I am a bit surprised that this would not be flagged anyway. Maybe we could just alert to the user to the situation instead and have them fix this?

I do have concerns in general about increasing the amount of time its going to take, especially now that we are like 50% of the way through COP.

jason-p-pickering:
Looks good. Merging into dev.

PR #275, opened 2021-04-21
exportSubnatToDATIM change column name and fix issue with d$datim sub…

sam-bao:
…nat impatt subtables


jacksonsj (approved):
@sam-bao - This will work for now. I've scoped out a more complete alternative we can look at once this fire is out.

PR #283, opened 2021-04-27
DP-229: Improper use of dedupe mechs in SNUxIM tab

jason-p-pickering:
https://jira.pepfar.net/browse/DP-229


jacksonsj (changes requested):
Only change I would make here @jason-p-pickering would be to not also look for `DSD|TA`. If it has `0000[01]` that alone is a flag, even if it doesn't have `DSD|TA`.

PR #286, opened 2021-04-29
Update and create functions for producing test datasets from datapackr

jacksonsj:
Contains two functions to allow production of test data:
- createTestDataset produces a test dataset for a given country that crosses all possible PSNUs, mechs, DEs, and COCs.
- createTestDataset.holistic takes a more nuanced approach, producing a list object with various test datasets each aimed at testing a specific data management assumption involved in COP.


NAMESPACE
```diff
@@ -34,6 +34,7 @@ export(countryUIDs_homeCell)
 export(createAnalytics)
 export(createKeychainInfo)
 export(createTestDataset)
+export(createTestDataset.holistic)
```
jason-p-pickering, line 37: Could we avoid the use of "." in function names? This is really just more of a code style issue I think. 

Perhaps we could have `createTestDataSet(....,type="holistic")` as a wrapper to various types of test datasets?

R/createTestDataset.holistic.R
```diff
@@ -0,0 +1,283 @@
...
+      value = sample(0:8, dplyr::n(), replace = TRUE)
+    ) %>%
+    dplyr::select(dataElement,
+                  period,
+                  orgUnit,
+                  categoryOptionCombo,
+                  attributeOptionCombo,
+                  value)
+  
+  #TODO: Functionalize repetitive code above
```
jason-p-pickering, line 270: Yes, I would agree with this. Should be fairly simple ?

jason-p-pickering (changes requested):
In general I am OK with this,but please consider the code style changes which I have suggested before merging.

cnemarich:
Too far out of sync with `dev`. Cherry-picked relevant commits on `update-createTestDataset-repair`. Please refer back to these notes from @jason-p-pickering on suggested updates.

PR #288, opened 2021-05-05
datapackVsDATIM start including dedupes in comparison

sam-bao:
Begin including dedupes in the psnuXim comparison. Change is required as we now import dedupes from datapacks

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: datapackr
 Type: Package
 Title: A Package that Packs and Unpacks all Data Packs
-Version: 4.5.0
+Version: 4.5.1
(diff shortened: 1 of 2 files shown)


jason-p-pickering (commented):
Looks in general good. Only question here is does this change work for COP20 OPUs or do we need to add some conditional logic here based on the COP year?

sam-bao:
This won't affect COP20 OPUs because there is a separate comparison function for that. I will check after COP and see if they can be combined. They originally diverged because we started including dedupes in COP20 OPU datapacks, since we now include these in the COP21 datapacks as well, we can probably reduce this to one function for both cases.

PR #289, opened 2021-05-05
Resolves DP-238 and DP-237

jacksonsj:
- Updated template with new formula for TX_TB_PREV tab, col K (TX_CURR FY21): to reference Cascade!$U:$U instead of Cascade!$AE:$AE.
- Updated template with new formula for HTS tab, col Z (TB_STAT New Positive %): to reference HTS!$K:$K instead of HTS!$J:$J
- Updated template schema object to reflect new formulas for validation purposes.

@jason-p-pickering or @sam-bao - Can you review?


jacksonsj:
Approved by @jason-p-pickering per Slack comments

PR #291, opened 2021-05-18
Addresses DP-250

jacksonsj:
- Adds flag for d$info$needs_psnuxim in both OPU and COP unpacking processes

### DIFF
--- a/R/unPackOPU_PSNUxIM.R
+++ b/R/unPackOPU_PSNUxIM.R
@@ -40,7 +40,6 @@ unPackOPU_PSNUxIM <- function(d) {
   
   if (NROW(d$data$extract) == 1 & is.na(d$data$extract[[1,1]])) {
     d$info$has_psnuxim <- FALSE
-    d$info$needs_psnuxim <- TRUE
     d$info$has_error <- TRUE
     
     warning_msg <- 
--- a/R/unPackOPU_PSNUxIM.R
+++ b/R/unPackOPU_PSNUxIM.R
@@ -38,6 +38,23 @@ unPackOPU_PSNUxIM <- function(d) {
       .name_repair = "minimal"
     )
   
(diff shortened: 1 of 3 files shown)


R/unPackOPU_PSNUxIM.R
```diff
@@ -38,6 +38,23 @@ unPackOPU_PSNUxIM <- function(d) {
       .name_repair = "minimal"
     )
   
+  if (NROW(d$data$extract) == 1 & is.na(d$data$extract[[1,1]])) {
+    d$info$has_psnuxim <- FALSE
+    d$info$needs_psnuxim <- TRUE
```
jason-p-pickering, line 43: Didn't we say that OPUs did not need a PSNUxIM?

jacksonsj: True, but this is checking for an extreme error case where, for some reason, a user submits a completely empty OPU Data Pack PSUxIM tab. In which case, it's both a major error and, technically, they need a PSUxIM tab.

jason-p-pickering: OK, this may have unintended consequences then, as I am planning on using this flag to control whether the option for "New PSNUxIM" should be available in the downloads section of the app. I am fine with keeping it, but then we will need functionality to allow people to populate an OPU from the app. Otherwise, I would need to add some special handling here, which I think is what we are trying to avoid.

jacksonsj: K, I've removed that piece to turn on the needs_psnuxim flag for empty OPU Data Packs. The error flag and error message should get the job done without gumming up the other parts you're working on.

jason-p-pickering (commented):
Look good. Have merged into unit-testing-redux which we can then merge into dev. Thus, closing this PR.

PR #293, opened 2021-05-21
Issues/dp235

jacksonsj:
@jason-p-pickering - This code:

1. Drops the automation for dragging down right side formulas on the PSNUxIM tab, reducing processing time significantly.
2. Adds a test for where formulas are missing on the right side.


jason-p-pickering (commented):
Looks good. I have merged into unit-testing-redux, which I think we can then merge into dev, so will close this PR.

PR #294, opened 2021-05-27
Address DP-255

jacksonsj:
Address Jira/DP-255


jason-p-pickering (approved):
I tested this and seems to work, so going to merge. Will add tests in the unit-testing-redux branch.

PR #295, opened 2021-05-28
DP-259

sam-bao:
Modify getOPUDataFromDATIM to explicitly request mer_targets when calling getCOPDataFromDATIM


jason-p-pickering (approved):
Smoked tested it by generating an OPU datapack for Lesotho. Seems to work.

PR #297, opened 2021-06-25
add mock for getDataValueSets test

sam-bao:
Adds mock and related content for getDataValueSets


sam-bao:
@jason-p-pickering this pull request adds a mock for the getDataValueSets test.

Additionally I removed the datimutils login call. Rather than creating the login object (even mocked) on the fly for each test, I added a stub for the required login object in the helper.R file. You can see that the CircleCI tests still pass after this change. I think this is a bit cleaner and allows each test to focus on testing the functions it is supposed to be testing without extraneous overhead.

If you agree with this approach I will continue to remove/replace an login function calls that are not meant to explicitly test logging in.

jason-p-pickering:
Looks good. Merging this one.

PR #300, opened 2021-07-02
Test compare schema and template

sam-bao:
Adds tests for comparing results of unpack schema with schema RDAs


jason-p-pickering (commented):
Looks good. Merging.

PR #302, opened 2021-08-17
Issues/dp 281 - COP21 OPU Data Pack generation

jacksonsj:


jacksonsj:
@sam-bao - Would you mind giving this a quick review before we pull into `dev`? We'll do a subsequent, more comprehensive review here soon of `dev` -> `master`.

R/adorn_import_file.R
```diff
@@ -118,43 +121,48 @@ adorn_import_file <- function(psnu_import_file,
...
-                   sex = Sex,
-                   key_population = KeyPop,
-                   resultstatus_specific = resultstatus,
-                   upload_timestamp,
-                   disagg_type,
-                   resultstatus_inclusive,
-                   top_level,
-                   target_value = value,
-                   indicator_code)
+  if (filter_rename_output) {
```
jacksonsj, line 129: @sam-bao - This change uses a parameter to allow the default to remain as it was for all existing functions that used this, but adds an option I use for COP21 OPU creation to keep all columns in the output and filter more selectively within a downstream function.

R/create_play_spectrum_output.R
```diff
@@ -186,7 +186,7 @@ create_play_spectrum_output <- function(country_uids,
                                             d2_session = d2_session)
     exportPackr(data = play_spectrum_output,
                 output_path = output_folder,
-                type = "Spectrum Example",
+                tool = "Spectrum Example",
```
jacksonsj, line 189: @sam-bao - As part of this branch, I reviewed some of the parameters used across OPU generation and saw that we use `type` and `tool` interchangeably, but also use `type` as a parameter in other functions though referring to something entirely different from whether a COP or OPU Data Pack. So I standardize the parameter names we use across these functions to use only `tool` to describe whether a process is producing/validating a COP or OPU Data Pack.

R/getOPUDataFromDATIM.R
```diff
@@ -83,19 +83,32 @@ getOPUDataFromDATIM <- function(cop_year,
...
-                  Age = valid_ages.name,
-                  sex_option_uid = valid_sexes.id,
-                  Sex = valid_sexes.name,
-                  kp_option_uid = valid_kps.id,
-                  KeyPop = valid_kps.name,
-                  attribute_option = attributeOptionCombo,
-                  value)
+    dplyr::filter(indicator_code %in% indicator_codes)
+  
+  # COP21+: Output as DHIS2 import file ####
```
jacksonsj, line 88: @sam-bao - This keeps the output identical for existing functions, but for COP21 OPU Data Packs, produces the output as a DHIS2 import file. We could also structure this to use a parameter option instead — e.g., `import_structure = TRUE` turns on this feature, but the default is `= FALSE` to allow backward compatibility with all the places using this function already.

R/packPSNUxIM.R
```diff
@@ -0,0 +1,608 @@
...
+#'
+#' @description Packs the PSNUxIM tab in either a COP or OPU Data Pack.
+#'
+#' @param data Dataset containing totals for allocation within PSNUxIM tab,
+#' formatted as a standard DHIS2 import file.
+#' @inheritParams datapackr_params
+#' 
+#' @return r Sidecar object containing both an openxlsx Workbook and alert messages
+#' 
+packPSNUxIM <- function(wb,
```
jacksonsj, line 13: @sam-bao - This function will eventually entirely replace all the various ones we had before for packing the differently structure PSNUxIM tabs we had between COP and OPU and 2019 and 2020 and 2021. For now, I've set this up to only work for COP21 OPU Data Pack PSNUxIM tabs, but most of the code comes from the `packSNUxIM` function we used for COP21 Data Pack PSNUxIM tabs, anticipating eventually using this one function to write either COP or OPU PSNUxIM tabs.

R/packSNUxIM_OPU.R
```diff
@@ -14,12 +14,12 @@ packSNUxIM_OPU <- function(d) {
     stop("Sorry! We're only set up to run this for COP20 OPUs for right now. Check back later please. Stay safe!")
   }
   
-  if (NROW(d$data$model_data) == 0) {
-    warning("No model data supplied.")
+  if (NROW(d$data$snuxim_model_data) == 0) {
```
jacksonsj, line 17: @sam-bao - Just standardizing here and elsewhere how we refer to the SNUxIM model data, to distinguish from the Data Pack Model Data we use in first populating COP Data Packs

R/packageSetup.R
```diff
@@ -75,3 +75,423 @@ headerRow <- function(tool, cop_year = getCurrentCOPYear()) {
...
+#' @param schema Which datapackr schema to use in guiding this function. If left
+#' \code{NULL} will select the default based on \code{cop_year} and \code{tool}.
+#' @param wb Openxlsx workbook object.
+#' @param PSNUs Dataframe of PSNUs to use in this function, containing at least
+#' \code{psnu_uid}.
+#' @param tool Type of tool this function will create or interact with. Either
+#' \code{OPU Data Pack} or \code{Data Pack}
+#' @param season Either \code{COP} or \code{OPU}.
+#'
+datapackr_params <- function(model_data,
```
jacksonsj, line 187: @sam-bao - Eventually, we can use this function and @inheritParams in `roxygyen` to keep just one central, standardized list of parameter definitions we can then reference in other functions where we want to. Again, this came up as an idea after reviewing our functions and realizing how disjointed our use of parameter names and meanings was across the package.

R/packageSetup.R
```diff
@@ -75,3 +75,423 @@ headerRow <- function(tool, cop_year = getCurrentCOPYear()) {
...
+}
+
+#' @export
+#' @title Standardized package function parameter definitions
+#' 
+#' @inheritParams datapackr_params
+#'
+#' @return params List of valid parameters to use in functions.
+#'
+check_params <- function(country_uids,
```
jacksonsj, line 215: @sam-bao - This can grow with our package and for now only really covers usage within the OPU Data Pack realm, but the idea is to standardize how we validate supplied parameters and fill in unsupplied/`NULL` ones.

R/utilities.R
```diff
@@ -535,4 +535,109 @@ getDataPackSchema <- function(cop_year) {
...
+
+#' @export
+#' @title Create a clean Data Pack Workbook shell
+#' 
+#' @param tool Either "Data Pack" or "OPU Data Pack"? Default is "Data Pack". 
+#' @inheritParams datapackr_params
+#'
+#' @return wb Data Pack Workbook shell
+#'
+createWorkbook <- function(datapack_name = NULL,
```
jacksonsj, line 550: @sam-bao - This is part of the process of standardizing how we fill in unsupplied/`NULL` parameters, so if we wanted to pull a function out and run it more independently, we could quick create a shell Workbook based on the supplied COP Year and tool type to then use within a function that needs a Workbook.

R/packOPUDataPack.R
```diff
@@ -1,31 +1,17 @@
...
-#' \code{NULL}, will prompt for file selection via window.
-#' @param  cop_year Specifies COP year for dating as well as selection of
-#' templates.
-#' @param output_folder Local folder where you would like your Data Pack to be
-#' saved upon export. If left as \code{NULL}, will output to
-#' \code{Working Directory}.
-#' @param results_archive If TRUE, will export compiled results of all tests and
-#' processes to output_folder.
-#' @param d2_session DHIS2 Session id
+#' @inheritParams datapackr_params
```
jacksonsj, line 9: @sam-bao - See comment in `packageSetup`

R/packOPUDataPack.R
```diff
@@ -47,86 +29,138 @@ packOPUDataPack <- function(datapack_name,
...
+                                             d2_session = d2_session)
+    if (NROW(d$data$snuxim_model_data) == 0) {
+      stop("SNUxIM Model data pull seems to have returned no data from DATIM. Please check with DATIM.")
+    }
+  }
   
-  if (NROW(d$data$model_data) == 0) {
-    stop("Model data pull seems to have returned no data. Please check with DATIM.")
+  # Prepare totals data for allocation ####
+  if (d$info$cop_year == 2021) {
```
jacksonsj, line 71: @sam-bao - For COP21 OPU Data Packs, we'll need to separate and `summarise` the target totals and use these within `packPSNUxIM`, rather than using a function to pull these targets from other tabs in the Data Pack. This is part of trying to design `packPSNUxIM` to be compatible with both COP and OPU PSNUxIM tabs.

R/packOPUDataPack.R
```diff
@@ -47,86 +29,138 @@ packOPUDataPack <- function(datapack_name,
...
       add_dp_psnu(.) %>%
       dplyr::arrange(dp_psnu) %>%
       dplyr::select(PSNU = dp_psnu, psnu_uid)
     
     # Write PSNUxIM tab ####
-    d <- packSNUxIM_OPU(d)
+    if (d$info$cop_year == 2020) {
+      d <- packSNUxIM_OPU(d)
+    } else {
+      r <- packPSNUxIM(wb = d$tool$wb,
```
jacksonsj, line 112: I've written `packPSNUxIM` to return a small list object called `r` that has only two components: the final workbook with populated PSNUxIM tab, and any messages. This same method could be used across the package for other functions to move away from monolithic and opaque usage of the `d` object where it's unclear for any given function what that needs to include, to supplying these pieces more explicitly as parameters and returning both the outputs and messages as a separate list object to then be reassigned into `d` as needed. This should also help with testability and modularization.

R/packOPUDataPack.R
```diff
@@ -47,86 +29,138 @@ packOPUDataPack <- function(datapack_name,
...
+      
+      d$tool$wb <- r$wb
+      d$info$warning_msg %<>% append(r$messages)
+    }
     
     # Save & Export Workbook
     print("Saving...")
     exportPackr(data = d$tool$wb,
                 output_path = d$keychain$output_folder,
-                type = d$info$type,
```
jason-p-pickering, line 120: Seems to be some confusion here as to whether use `tool` or `type`. If we change this from d$info$tool to d$info$type, that is going to be a breaking change with the self-service app. We can of course change it there, but is this really a necessary change?

jason-p-pickering (changes requested):
Please see my comment below regarding the change and/or confusion around d$info$tool.

R/packageSetup.R
```diff
@@ -75,3 +75,423 @@ headerRow <- function(tool, cop_year = getCurrentCOPYear()) {
...
+#' @export
+#' @title Pick correct schema
+#' 
+#' @param cop_year Specifies COP year for dating as well as selection of
+#' templates.
+#' @param tool Either Data Pack or OPU Data Pack
+#'
+#' @return Schema file for given cop_year and tool type
+#'
+pick_schema <- function(cop_year = getCurrentCOPYear(), tool = "Data Pack") {
```
sam-bao, line 88: @jacksonsj this partially overlaps with this function: https://github.com/pepfar-datim/datapackr/blob/ed4f0a0bddecc430b9fa84230a24f08c109fef3c/R/utilities.R#L530. 

Should we consolidate and update the code which utilizes getDataPackSchema

sam-bao: Unless there is a reason for us to call this function regularly from the commandline I don't think it should have a default cop_year. It should receive cop_year explicitly from a higher level function like packOPUDataPack which may have a default cop_year. The risk is that we write a higher level function which uses this function, we forget to explicitly pass cop_year when that higher level function calls `pick_schema` and then it uses a default schema for the current cop year which could be inconsistent with the higher level function. So leaving the default for cop_year off of these utility functions is more defensive programming.

jacksonsj: Agreed. I'll change.

jacksonsj: @sam-bao - I've addressed this here: https://github.com/pepfar-datim/datapackr/pull/302/commits/8b2733cdc7bbe78c4af0e25d8e8af53eacf93bb5

R/packOPUDataPack.R
jacksonsj: Debatable, for sure. I think overall, we definitely have issues with lack of consistency around our use of parameter names and definitions. We'll need to solve that at some point.

R/packageSetup.R
sam-bao: @jacksonsj we still have the redundant function `pick_schema` and `getDataPackSchema`. I would probably just convert pick_schema to getOPUDataPackSchema(cop_year)

jason-p-pickering: Or maybe just `getSchema(tool = "OPU Data Pack", cop_year=2021)`. I am not totally sure why we need to distinguish in the function name? It would seem that both an OPU and DataPack have a schema and we just need to tell the function which type of tool we are dealing with?

R/checkNumericValues.R
```diff
@@ -72,7 +72,7 @@ checkNumericValues <- function(d, sheet, header_cols = NULL){
         "WARNING! In tab ",
         sheet,
         ": NON-NUMERIC VALUES found! Please ensure all values entered against",
-        " FY22 Target columns include numeric values only — no letters or punctuation.",
+        " FY22 Target columns include numeric values only - no letters or punctuation.",
```
jason-p-pickering, line 75: BTW @sam-bao and @jacksonsj , the em-dash (long) is a non ASCII character, and the package checks complained about this. This was the reason for this change.

R/handshakeFile.R
```diff
@@ -38,8 +38,10 @@ canReadFile <- function(path) {
 #' @return Character vector containing valid filepath for further use.
 #'
 handshakeFile <- function(path,
-                          tool) {
+                          tool = "Data Pack") {
```
sam-bao, line 41: @jacksonsj I don't see why this needs a default value. Shouldn't we be explicit in the tool we are getting?

R/packDataPack.R
```diff
@@ -56,7 +57,7 @@ packDataPack <- function(model_data,
   )
 
   # Open schema ####
-  if (cop_year == 2020) {
+  if (d$info$cop_year == 2020) {
```
sam-bao, line 60: @jacksonsj doesn't the pick_schema funtion do this?

R/packageSetup.R
jason-p-pickering: When writing tests, I came across this function. It is odd in that it does nothing but return NULL, which itself is not explicitly declared. This is generally not good practice. I think the better approach would be to have this function return the list of formal arguments as a list, which we can then use as a test for the other functions which you have created. 

so, if it datapackr_params returned `rlang::fn_fmls_names(fn = datapackr_params)` then we know exactly which arguments are being called (well at least their names) and can then test the other functions which you have implemented using this constructor. 

Thoughts @sam-bao @jacksonsj ?

R/prepare_model_data.PSNUxIM.R
```diff
@@ -1,47 +1,75 @@
...
-prepare_model_data.PSNUxIM <- function(model_data,
+prepare_model_data.PSNUxIM <- function(snuxim_model_data,
                                        country_uids) {
   
   interactive_print("Getting data about your FY21 Mechanism Allocations from DATIM...")
   
   # Pivot wider ####
-  snuxim_model_data <- model_data[country_uids] %>%
-    dplyr::bind_rows() %>%
+  if (inherits(snuxim_model_data, "list")) {
```
sam-bao, line 18: @jacksonsj I don't really understand what this if statement is for, can you explain what is going on here?

R/prepare_model_data.PSNUxIM.R
```diff
@@ -1,47 +1,75 @@
...
+    snuxim_model_data <- snuxim_model_data[country_uids] %>%
+      dplyr::bind_rows()
+  }
+  
+  # Drop all data that can't be allocated across mech & DSD/TA ####
+  snuxim_model_data %<>%
+    dplyr::filter(stringr::str_detect(mechanism_code, "\\d{4,}"),
```
sam-bao, line 25: @jacksonsj just so I am clear - is this mainly to drop DREAMS data at the moment?

R/prepare_model_data.PSNUxIM.R
```diff
@@ -50,7 +78,7 @@ prepare_model_data.PSNUxIM <- function(model_data,
   # Create Deduplicated Rollups ####
   snuxim_model_data %<>%
     dplyr::mutate(
-      `Total Duplicated Rollup` = rowSums(dplyr::select(., tidyselect::matches("\\d{4,}")), na.rm = TRUE),
+      `Total Duplicated Rollup` = rowSums(dplyr::select(., tidyselect::matches("\\d{4,}|HllvX50cXC0")), na.rm = TRUE),
```
sam-bao, line 81: @jacksonsj If we drop dreams data, is the default mechanism supposed to be listed here?

R/packageSetup.R
sam-bao: @jacksonsj a bit confused, there are a number of functions nested in this function but then they aren't actually called. Is this functional? If not I say we pull this into a separate feature branch. I think this needs unit tests before it is merged as well.

R/packageSetup.R
```diff
@@ -75,3 +75,421 @@ headerRow <- function(tool, cop_year = getCurrentCOPYear()) {
...
+#' @return Schema file for given cop_year and tool type
+#'
+pick_schema <- function(cop_year, tool) {
+  
+  if (tool == "OPU Data Pack") {
+    if (cop_year == 2020) {
+      schema <- datapackr::cop20OPU_data_pack_schema
+    } else if (cop_year == 2021) {
+      schema <- datapackr::cop21OPU_data_pack_schema}  
+  }
```
sam-bao, line 95: @jacksonsj what if not 2020 or 2021?

R/packageSetup.R
```diff
@@ -75,3 +75,421 @@ headerRow <- function(tool, cop_year = getCurrentCOPYear()) {
...
+  
+  if (tool == "Data Pack") {
+    if (cop_year == 2020) {
+      schema <- datapackr::cop20_data_pack_schema
+    } else if (cop_year == 2021) {
+      schema <- datapackr::cop21_data_pack_schema
+    } else {schema <- datapackr::data_pack_schema}
```
sam-bao, line 102: Not sure I like that if I give a cop year of 2022 right now that I get the cop21 schema. In theory we shouldn't call this with 2022 until there is a COP22 schema - but new team members might be confused. In fact I think the cop22 schema should start its life as the cop22 schema not as the data_pack_schema.

R/packageSetup.R
jason-p-pickering: I started with a few unit tests, and agree, that we need tests here. I do not see a compelling reason to have these functions nested. I think they can simply not be exported, if they are part of something bigger, but as separate functions, it makes them much easier to individually test. See some of my tests in 4eb5d224f03979ce8d383aa8b7c8c8f4dd5840e4, which I think could be extended to provide more coverage.

R/prepare_model_data.PSNUxIM.R
jason-p-pickering: For me the bigger question here is what happens when its not a list or its a list but with the wrong structure. I think it would be better to be explicit and be sure that whatever is being passed via the `snuxim_model_data` is what we actually expect it is and not simply a list.

R/packOPUDataPack.R
```diff
@@ -47,86 +29,138 @@ packOPUDataPack <- function(datapack_name,
...
+  }
   
-  if (NROW(d$data$model_data) == 0) {
-    stop("Model data pull seems to have returned no data. Please check with DATIM.")
+  # Prepare totals data for allocation ####
+  if (d$info$cop_year == 2021) {
+    d$data$UndistributedMER <- d$data$snuxim_model_data %>%
+        dplyr::mutate(attributeOptionCombo = default_catOptCombo()) %>%
+        dplyr::group_by(dplyr::across(c(-value))) %>%
+        dplyr::summarise(value = sum(value)) %>%
```
sam-bao, line 75: @jacksonsj any chance for `NA` here? If so the sum will be `NA` unles we set na.rm=TRUE

R/packageSetup.R
jason-p-pickering: See my comment in the tests, where this did not feel entirely right to me. 

```
  #Are we sure we want to return this object for anything other than 2020 
  # and 2021?
  test_schema<-pick_schema(1999,"Data Pack") 
  testthat::expect_identical(test_schema,  datapackr::data_pack_schema)

File: R/packOPUDataPack.R [hunk 23]
```diff
@@ -47,86 +29,138 @@ packOPUDataPack <- function(datapack_name,
...
+  
+  if (!d$info$cop_year %in% c(2020, 2021)) {
+    stop("Sorry! We're only set up to run this for COP20 or COP21 OPUs.")
+  }
+  
+  # Check if provided model data is empty ####
+  if (!is.null(d$data$snuxim_model_data)) {
+    
+    empty_snuxim_model_data <- d$data$snuxim_model_data %>%
+      dplyr::filter(rowSums(is.na(.)) != ncol(.))
```
sam-bao, line 52: @jacksonsj  I had trouble understanding the logic here. @jason-p-pickering helped me unpack it, and he suggested this as perhaps a more straightforward check.

`all(is.na(bar))`

Jason's suggestion seems easier to understand if the intent is to check for a dataframe with all NA.

NAMESPACE
```diff
@@ -28,16 +28,20 @@ export(checkInvalidOrgUnits)
 export(checkMissingMetadata)
 export(checkNumericValues)
 export(checkStructure)
+export(check_params)
```
sam-bao, line 31: @jacksonsj Shouldn't this be called `checkParams`? There are other examples of function names here that diverge from camelCase as well

R/packageSetup.R
sam-bao: I am also looking at the usage of check_params in packPSNUxIM, and I now see the thinking is that this will not just check the validity of parameters but also transform them and pass them back to the calling function. I am not sure letting check_params alter the value of parameters is a good thing. Especially when it is happening in a function like `packPSNUxIM` which we will not call directly. parameters set in a higher level function shouldn't have the chance to get altered in internal function like `packPSNUxIM`. 

Also to reiterate a point I think I made before, I don't think we should be exporting every function in the package. Other than for development and testing `packPSNUxIM` will only be called by another `datapackr` function. So if there was a need to transform a parameter it should have happened before that parameter got to packPSNUxIM.

To refocus on the issue at hand here, I am really nervous that check_params is potential altering values of parameters in low level functions, and even in higher level functions may cause confusion.

R/packageSetup.R
jacksonsj: I'll create a ticket for consolidating, but would like to handle in a separate PR: https://jira.pepfar.net/browse/DP-289

R/handshakeFile.R
jacksonsj: I'll create a ticket to analyze and standardize how we handle parameter defaults across the package: https://jira.pepfar.net/browse/DP-290

NAMESPACE
jacksonsj: Documenting and will resolve via: https://jira.pepfar.net/browse/DP-291

R/packageSetup.R
jacksonsj: Will resolve via: https://jira.pepfar.net/browse/DP-290

R/prepare_model_data.PSNUxIM.R
sam-bao: can we us `is.list()`?

sam-bao: or `typeof`

R/prepare_model_data.PSNUxIM.R
jacksonsj: Yes, but rather than explicitly drop `AGYW_PREV`, I've generalized.

R/prepare_model_data.PSNUxIM.R
jacksonsj: Yes, probably could remove.

sam-bao: Probably redundant, @sam-bao to double check and remove.

R/prepare_model_data.PSNUxIM.R
jacksonsj: I'll review and see if we can make this more explicit.

R/packageSetup.R
jacksonsj: Will resolve this here: https://jira.pepfar.net/browse/DP-292

R/packageSetup.R
jacksonsj: Will resolve this here: https://jira.pepfar.net/browse/DP-292

R/packOPUDataPack.R
jacksonsj: I'll look at this and try to resolve in this PR.

R/packOPUDataPack.R
jacksonsj: In other words, yes, this needs to remain, and we'll have to update the SSA code.

R/packDataPack.R
jacksonsj: Yes, but I felt like changing that here was out of scope for this PR since this is focused on OPU generation, not COP generation. I'll create a new ticket for this and resolve in a separate PR.

https://jira.pepfar.net/browse/DP-294

R/utilities.R
jacksonsj: Noting here that I'll need to change this function name to deconflict with `openxlsx`: https://jira.pepfar.net/browse/DP-295

sam-bao (approved):
@jacksonsj I am done with QA. I think there were a few minor changes you wanted to add to this PR, so rather than merging I am marking as approved for you to merge when you are done.

Should we increment to version 5.0.0? There are some breaking changes so that would seem appropriate.

PR #306, opened 2021-08-31
Update valid_PSNUs.rda with COP20 WAR PSNUs

sam-bao:
The updated valid PSNUs file include Sierra Leone and Burkina Faso.


jason-p-pickering (commented):
Were these manually added to the object? I would have expected some changes [here](https://github.com/pepfar-datim/datapackr/blob/master/R/getPSNUs.R)

sam-bao:
@jason-p-pickering yes, I added them by adding Sierra Leone and Burkina to the PSNU org unit group in triage and running the update script. I ensured only these two org units were added and none were deleted from the valid PSNU file.

We could add something to that script I suppose so these org units don't get deleted when we update that file for COP22, but I think we all agreed that we still need to implement a more general solution before then anyway.

I will take a look and see if I can also add those two countries directly in the script.

sam-bao:
@jason-p-pickering can you still proceed to make sure that the updated file allows the app to function correctly. I confimed the d$data$analytics data look OK, but am not sure if the app relies on the valid_psnu file in other ways.

sam-bao:
@jason-p-pickering rather than modifying the getPSNUs function to return Sierre Leone and Burkina Faso, I instead modified the script we use to create valid_psnus to prevent deleting those two "PSNUs".

PR #308, opened 2021-09-01
Message refactor

jason-p-pickering:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

- Refactor of message queue to use a simple R6 object instead of a list

## Affected Process Elements
- [ ] Data Pack generation
- [ ] Data Pack model
- [ ] Data Pack self-service app
- [ ] Data Pack processing
- [ ] Site Tool generation (from DATIM - for OPUs)
- [ ] Site Tool processing
- [ ] Site Tool import

## Types of changes

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes)
- [ ] Build-related changes

- [X] Yes
- [ ] No

This will break the self-service apps for sure, but should be easy to rectify this later. 

## Final Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules


R/messageQueue.R
```diff
@@ -0,0 +1,85 @@
...
+  ## Set the name for the class
+  class(me)<-c("data.frame","MessageQueue")
+  return(me)
+}
+
+
+appendMessage<-function(x,message,level) {
+  UseMethod("appendMessage",x) }
+
+appendMessage.MessageQueue<-function(x, message=NA,level=NA) {
```
jacksonsj, line 28: Should any of these here be exported, or at least declared as package functions for reference?

R/messageQueue.R
```diff
@@ -0,0 +1,85 @@
...
+      "KEY:\r\n",
+      "- WARNING!: Problematic, but doesn't stop us from processing your tool. May waive with approval from PPM and DUIT.\r\n",
+      "- ERROR!: You MUST address these issues and resubmit your tool.\r\n",
+      "*********************\r\n\r\n")
+    
+    cat(crayon::red(crayon::bold("VALIDATION ISSUES: \r\n\r\n")))
+    cat(crayon::red(key))
+    cat(crayon::red(messages))
+  }
+}
```
jacksonsj: Has this been `lintr`ed yet?

R/packPSNUxIM.R
```diff
@@ -600,7 +600,7 @@ packPSNUxIM <- function(wb,
       "If you have any questions, please submit a Help Desk ticket at DATIM.Zendesk.com.",
       "\n")
   
-  r$messages <- append(r$messages, warning_msg)
+  r$messages$append(warning_msg,"INFO")
```
jacksonsj: Not sure I understand how this change works. Should this be `appendMessage` instead of `append`?

R/unPackOPU_PSNUxIM.R
```diff
@@ -308,7 +308,8 @@ unPackOPU_PSNUxIM <- function(d) {
           collapse = "\n\t* "),
         "\n")
     
-    d$info$warning_msg <- append(d$info$warning_msg, warning_msg)
+    d$info$messages <- appendMessage(d$info$messages, warning_msg,"ERROR")
+    d$info$has_error<-TRUE
```
jacksonsj, line 314: Instead, could we trigger this `has_error` flag automatically at the end just by the presence of a message with the `"ERROR"` level?

jacksonsj (changes requested):

R/messageQueue.R
jason-p-pickering: I have exported the `printMessages` method. I think that is really the only one which needs to be exposed.

R/messageQueue.R
jason-p-pickering: Not sure what you mean, but yes. :-D

R/packPSNUxIM.R
jason-p-pickering: Good catch. I had done a search/replace on `d$info$messages` but this object was structured slightly differently for some reason. I made two changes [here](https://github.com/pepfar-datim/datapackr/blob/d1adb81a88b413739b85e3bfcd542850c180d538/R/packPSNUxIM.R#L42) and [here](https://github.com/pepfar-datim/datapackr/blob/d1adb81a88b413739b85e3bfcd542850c180d538/R/packPSNUxIM.R#L599). This method seems a bit odd to me. Since it kind of constructs a `d` like object, but did not have the same structure. I moved things around a bit to make it more standard, but we might need to take a look at this in more depth to be sure I did not break anything unintentionally.

R/unPackOPU_PSNUxIM.R
jason-p-pickering: Yes, I had considered doing this. It should be pretty easy to implement, but would suggest we perform another refactor after all of this is merged.

R/checkColStructure.R
```diff
@@ -7,50 +7,47 @@
...
+        ~ .)
+
   col_check <- d$info$schema %>%
     dplyr::filter(sheet_name == sheet) %>%
     dplyr::select(indicator_code, template_order = col) %>%
-    {if (sheet == "PSNUxIM") {
-      dplyr::filter(.,
-                    !stringr::str_detect(indicator_code, "\\d{4,}_(DSD|TA)|^$"))
-    } else {.}} %>%
+    purrr::when(sheet == "PSNUxIM" ~ dplyr::filter(., !stringr::str_detect(indicator_code, "\\d{4,}_(DSD|TA)|^$")),
```
jacksonsj, line 33: Bravo. Didn't know this existed.

jacksonsj (approved):

PR #340, opened 2021-11-18
Betaset data

jacksonsj:

- Adds execution script to produce test Datasets for use in COP22 testing.
- Adds COP22 map between Data Pack indicator codes and DATIM DEs/COCs
- Updates valid_PSNUs
- Adds execution script to extract Spectrum data from COP 21 Data Packs for use in producing Beta Packs
- Increments api_version()
- Updates DEGS ids for HTS modalities to accommodate new SNS modality

### Related Issues
- DP-354

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/prepareSheetData.R
```diff
@@ -88,8 +88,8 @@ prepareSheetData <- function(sheet,
       is.character,
       stringr::str_replace_all,
       pattern = paste0("(?<=[:upper:])", headerRow(tool = "Data Pack Template", cop_year = cop_year) + 1),
-      replacement = as.character(seq_along(NROW(row_headers))
```
jacksonsj, line 91: For some reason, this was causing all the formulas to repeat identically, rather than sequence along each row, so I reverted this one line.

R/pullFullCodeList.R
```diff
@@ -9,47 +9,57 @@
...
                                             "impatt"),
+                             datasets,
+                             expanded = FALSE,
                              d2_session = dynGet("d2_default_session",
-                                                 inherits = TRUE)
-                             ) {
+                                                 inherits = TRUE)) {
 
-  datasets_list <- character(0)
+  datasets <- datasets %missing% NULL
```
jacksonsj, line 27: This is the cause of the 1 failed test. It'll be fixed with that other PR, and adding this one function here would cause an annoying conflict. So, leaving this as is for now.

R/pullFullCodeList.R
```diff
@@ -9,47 +9,57 @@
...
-                       datapackr::getDatasetUids(FY, type = "subnat_results"))
-  }
-  if ("impatt" %in% datastream) {
-    datasets_list <- c(datasets_list,
-                       datapackr::getDatasetUids(FY, type = "impatt"))
-  }
-
-  datasets_list <- unique(datasets_list)
+  
+  datasets <- unique(datasets)
```
jacksonsj, line 56: All these changes basically just make this function more flexibly, to allow you to pass a list of datasets to grab instead of the whole custom selection thing that was there before (which still works if you want to use it).

R/pullHTSModality.R
```diff
@@ -30,13 +31,16 @@ getHTSModality <- function(cop_year = getCurrentCOPYear(), dataElements = NULL,
     dplyr::distinct() %>%
     dplyr::select(dataElement = id,
                   hts_modality = name) %>%
-    dplyr::mutate(hts_modality = stringr::str_trim(stringr::str_remove(hts_modality, fy_pattern)))
+    dplyr::mutate(
+      hts_modality =
+        stringr::str_trim(
+          stringr::str_remove_all(hts_modality, fy_pattern)))
```
jacksonsj, line 37: This and the fy_pattern update above are to accommodate how some of the newer groupSet names are named slightly differently. This pattern + str_remove_all catches all the varieties.

R/pullHTSModality.R
```diff
@@ -12,15 +12,16 @@
...
-  } else if (cop_year == 2019) {
-    groupSet <- "Jm6OwL9IqEa"
-  } else if (cop_year == 2018) {
-    groupSet <- "CKTkg8dLlr7"
-  }
-
-  fy_pattern <- "FY\\d{2},\\d{2}R/FY\\d{2},\\d{2}T"
+  groupSet <- switch(
+    as.character(cop_year),
+    "2022" = "a0oGIk4aN23",
```
jacksonsj, line 17: New groupSet created by Kyle for capturing the SNS modality

R/pullFullCodeList.R
jason-p-pickering: Feels repetitive. Can we simplify with `do.call`?

R/pullFullCodeList.R
jason-p-pickering: Once again, I am left scratching my head with the use of `%missing%` instead of just specifying `datasets=NULL` in the default paramaters?

R/packDataPack.R
```diff
@@ -17,27 +17,25 @@
...
 #' @return Exports a Data Pack to Excel within \code{output_folder}.
 #'
-#TODO Remove use of getwd()!!
 packDataPack <- function(model_data,
                          datapack_name,
                          country_uids,
                          template_path = NULL,
                          cop_year = getCurrentCOPYear(),
-                         output_folder = getwd(),
+                         output_folder,
```
jason-p-pickering, line 32: I think this needs some default parameter? We also need to check that the directory exists, and that we can write to it. If that condition is not met, then we should error out. 

Perhaps set the default to NULL and then 

```
  if ( is.null(output_folder) || file.access(output_folder,2) != 0 ) {
    stop("Cannot write to output_folder")
  }
```

R/packageSetup.R
```diff
@@ -322,7 +322,7 @@ check_params <- function(country_uids,
     if (is.null(cop_year)) {
       cop_year <- getCurrentCOPYear()
     }
-    if (!cop_year %in% c(2020, 2021)) {
+    if (!cop_year %in% c(2020, 2021, 2022)) {
       stop("Sorry, datapackr only supports COP20 and COP21 Data Packs.")
```
jason-p-pickering, line 326: This comment should be updated I think.

R/prepareSheetData.R
jason-p-pickering: It should be `seq_len(NROW(row_headers))`

jason-p-pickering (changes requested):
I will fix the linting, but please have look at my other comments before merging.

jason-p-pickering:
@jacksonsj unit tests are still failing due to the lack of availability of `%missing%`

R/packageSetup.R
jacksonsj: I've handled this in that other branch with full check_params implementation

R/pullFullCodeList.R
jacksonsj: Yes, but I'd rather do that in a separate PR focused just on that set of functions.

PR #372, opened 2022-01-26
DP-510

sam-bao:
Change compareData_OpuDatapackVsDatim to NOT compare AGYW data which are not part of OPU data packs.

### DIFF
--- a/R/datapackVsDatim.R
+++ b/R/datapackVsDatim.R
@@ -127,10 +127,10 @@ compareData_DatapackVsDatim <-
 # Get data from DATIM using data value sets
 
     datim_data <- dplyr::bind_rows(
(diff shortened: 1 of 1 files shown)


JordanBalesBAO (commented):
I like the solution!

I looked at 'assertthat' documentation on github and saw it will throw an error if the conditions isn't met. That is obviously the point, but this won't cripple any of our current downstream processes right?

sam-bao:
@JordanBalesBAO this function is called from the script that we use to import target updates "OPUs". So it shouldn't cause any issues elsewhere. That said if the assertion is violated it is a problem so we in fact would want it to throw an error down stream.

JordanBalesBAO (approved):

jason-p-pickering (approved):

PR #374, opened 2022-01-28
DP-517: Cleaning up warnings and notes during build checks

cnemarich:
Updates documentation, package dependencies, and global variables, as well as making minor bug fixes to clear all but two warnings and notes from the package build.
* Fixes minor issues in .Rbuildignore.
* Updates NAMESPACE file:
    - Downgrades R dependency from 4.1.1 to 4.1.0 (Clears note that R dependencies should be a "0" patch level).
    - Adds `datimutils` as a required package and adds associated remote.
    - Adds `waldo` as a suggested package.  
    - Moves `assertthat` from a required to a suggested package.
    - Removes `scales` as a required package as it is not used anywhere.
* Updates `data.R` file to fix mismatched bracket issue as well as add missing documentation for datasets in `data` folder.
* Changes `filepath` argument in `unPackDataPackTemplate` and `validateSchema` to `template_path` to be consistent with other functions and with documented parameters in `datapack_params`.
* Fixes mistyped function calls and argument names in several locations.
* Makes explicit which package certain functions are being called from.
* Adds missing global variables to `datapackr.R`.

Remaining notes include:
* The size of the package being 7.0 Mb with the `data` folder being 1.9Mb and the `extdata` folder being 4.1Mb. These issues should be cleared up by deleting objects related to COP19 and COP20 in the package (a different ticket).
* The number of hard package dependencies is at 27 non-default package. This should have a different ticket created for it.

### Related Issues
- DP-517

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


.Rbuildignore
```diff
@@ -3,17 +3,18 @@
...
+^\.idea
+^\.idea/.name
 ^packrat/
 ^\.Rprofile$
-\.github
-\.circleci
-\data-raw
+^\.github$
+^\.circleci$
+^data-raw$
```
jason-p-pickering, line 20: Are you sure we want to exclude this directory?

R/packSNUxIM_OPU.R
```diff
@@ -134,7 +134,7 @@ packSNUxIM_OPU <- function(d) {
     dplyr::select(col, indicator_code, column_names, formula)
 
   fx_cols <- schema_plus %>%
-    `row.names <- `(.[, "column_names"]) %>%
+    `row.names<-`(.[, "column_names"]) %>%
```
jason-p-pickering, line 137: Linting?

jason-p-pickering (changes requested):
Looks good I think, except for exclusion of the data-raw directory which I am not certain about.

R/packSNUxIM_OPU.R
cnemarich: Yeah, I think it still works properly with the spaces, but it shouldn't have them (and doesn't have them anywhere else this function is called).

.Rbuildignore
cnemarich: I was going off of this: https://r-pkgs.org/package-structure-state.html?q=data-raw#rbuildignore

And adding it clears these two notes and warnings:
```
❯ checking for portable file names ... WARNING
  Found the following files with non-portable file names:
    data-raw/~$COP21_Data_Pack_Template_example.xlsx
    data-raw/COP 2021 categoryOptionGroups.xml
    data-raw/COP 2021 categoryOptionGroupSet.xml
    data-raw/COP 2022 categoryOptionGroups.json
    data-raw/COP 2022 categoryOptionGroups.xml
    data-raw/COP 2022 categoryOptionGroupSet.json
    data-raw/COP 2022 categoryOptionGroupSet.xml
  These are not fully portable file names.
  See section ‘Package structure’ in the ‘Writing R Extensions’ manual.

❯ checking top-level files ... NOTE
  Non-standard file/directory found at top level:
    ‘data-raw’
```

jason-p-pickering: Yes, totally agree this is a problem with CRAN checks, and I see you have only put this in in .Rbuildignore. This seems correct to me, so fine to proceed.

R/packSNUxIM_OPU.R
jason-p-pickering: Hmm. Seems odd to me, as usually the linter complains if there is not a space around `<- `.

cnemarich: `row.names<-` seems to be a special case where it's all treated as one function rather than separate.
https://stat.ethz.ch/R-manual/R-devel/library/base/html/row.names.html

PR #377, opened 2022-01-31
Psnu parsing

jason-p-pickering:


R/unPackCountryUIDs.R
```diff
@@ -130,7 +115,39 @@ unPackCountryUIDs <- function(submission_path,
...
+  if (NROW(PSNUs) > 0) {
+    # TEST: Check country_uids and PSNUs in Data Pack match
+    if (!all(purrr::map_lgl(unique(PSNUs$country_uid),
+                            ~ .x %in% country_uids))) {
+      warning("Deduced or provided Country UIDs do no match Country UIDs observed in submission.")
+    }
+  } else {
+    warning("No PSNUs were detected.")
+  }
+
```
sam-bao, line 129: It looks like these checks on PSNU could be consolidated with the ones above (perhaps in the `if` block on line 55) requiring only one call to parsePSNU. Not required but I think makes the code a little cleaner.

sam-bao (approved):
One optional suggestion in the comments. 

Tested with the datapack referenced in DP-509 from the commandline and got expected error.

```
> d <- datapackr::unPackTool(file.choose())
Error in parsePSNUs(submission_path, tool, cop_year) :      
  ERROR: The PSNUxIM tab contains malformed PSNU identifiers. The following
      rows were affected:  Tanzania >Kinondoni MC . This error
      must be fixed in order to proceed.
```

PR #378, opened 2022-01-31
DP-504: Incorporate shared app functionality into package

cnemarich:
- Adds `source_user`, `sane_name`, `organisation_unit`, and `uuid` as metadata included in keychain.
    - Adds `d2_session` as an optional argument for the `createKeychainInfo` function to facilitate the creation of the `source_user` metadata.
- Adds `cached_mechs_path` as a standard parameter contained in the `datapackr_params` function.
- Remove the suggested library `doMC` and replace with `parallel`. If the library is present on the local system, indicator and validation rule evaluations will be run in parallel, usually resulting in a very significant performance enhancement. 
- Adds the following new functions to the package:
    - `checkMechanisms`, which checks that all mechanisms contained in the PSNUxIM tab are valid for the given COP year. This is based on the function `validateMechanisms` that was in `datapackr-app`.
    - `checkPSNUs`, which checks that all PSNUs contained in the Data Pack tool are contained within DATIM and match with the OU for the tool. This is based on the function `validatePSNUs` that was in `datapackr-app`.
    - `getOUFromCountryUIDs`, which fetches the parent OU name and UID for a tool. For single-country OUs, this simply returns the name and UID of the country, while for regional OUs, this function returns the name and UID of the parent region (e.g. Western Hemisphere). This is based on the function `getOrganisationUnitFromCountryUIDs` that was in `datapackr-app`.
    - `getSaneName` which takes a Data Pack tool name and generates a "Sane name" for the tool which has no spaces or punctuation.
    -  `fetchPrioritizationTable` retrieves an existing COP Memo prioritization table from DATIM to `d$data$memo$datim$prio` 
    -  `memoStructure` defines the COP memo structure as a DOCX output.
    -  `preparePartnerMemoTable` adds a data frame to ` d$data$partners_table` corresponding to the content of the partner memo table.
    - `preparePrioTable` adds a data frame to `d$data$prio_table` corresponding to the content of the prioritization table. 
    - `generateApprovalMemo` generates a COP memo in DOCX format.
    - `evaluateIndicators` function which evaluates a DHIS2 indicator expression and returns a numeric value. 

### Related Issues
- DP-504
- DP-536

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/checkPSNUData.R
```diff
@@ -0,0 +1,119 @@
...
+#'
+#' @description Tests PSNUxIM data
+#'
+#' @inheritParams datapackr_params
+#'
+#' @return d
+#'
+checkPSNUData  <-  function(d,
+                            validation_rules_path =
+                              paste0(Sys.getenv("support_files_directory"), "cop_validation_rules.rds"),
```
jason-p-pickering, line 12: Perhaps we should include the RDS of the validation rules in the package?

DESCRIPTION
```diff
@@ -59,7 +59,8 @@ Suggests:
     testthat,
     cellranger,
     rprojroot,
-    waldo
+    waldo,
```
jason-p-pickering, line 62: Are we sure this only needs to be a suggestion? Seems to be called only in `validateSchema`. Fine to leave it as a suggestion if never called by users.

R/checkMechanisms.R
```diff
@@ -0,0 +1,57 @@
...
+checkMechanisms <- function(d,
+                            cached_mechs_path = paste0(Sys.getenv("support_files_directory"), "mechs.rds"),
+                            d2_session = dynGet("d2_default_session",
+                                                inherits = TRUE)) {
+
+  mechs_data <- unique(d$datim$MER$attributeOptionCombo)
+
+  period_info <- datimvalidation::getPeriodFromISO(paste0(d$info$cop_year, "Oct"))
+
+  operating_unit <- getOperatingUnitFromCountryUIDs(d$info$country_uids)
```
jason-p-pickering, line 20: Don't we already the Operating Unit UID? If not, we should maybe take care of this when getting country UIDs?

R/checkMechanisms.R
```diff
@@ -0,0 +1,57 @@
...
+
+  bad_mechs <- mechs_data[!(mechs_data %in% mechs_datim)]
+
+  if (length(bad_mechs) > 0) {
+    msg <- paste0("ERROR!: Invalid mechanisms found in the PSNUxIM tab.
+                  These MUST be reallocated to a valid mechanism
+                  ", paste(bad_mechs, sep = "", collapse = ", "))
+    d$tests$bad_mechs <- bad_mechs
+    d$info$warning_msg <- appendMessage(msg, d$info$warning_msg)
+    d$info$had_error <- TRUE
```
jason-p-pickering, line 53: I think this should be `has_error`

R/checkPSNUData.R
```diff
@@ -0,0 +1,119 @@
...
+
+  # We need ALL mechanisms to be in DATIM before remapping....TODO
+  vr_data$attributeOptionCombo  <-
+    datimvalidation::remapMechs(vr_data$attributeOptionCombo,
+                                d2_session$user_orgunit,
+                                "code",
+                                "id",
+                                d2session = d2_session)
+
+  # TODO Can this be replaced with a call to `getDatasetUids`?
```
jason-p-pickering, line 56: Maybe we should take care of this while we are cleaning house?

R/createKeychainInfo.R
```diff
@@ -27,6 +22,11 @@ createKeychainInfo <- function(submission_path = NULL,
       cop_year = cop_year)
   )
 
+  # Pulls username if `d2_session` object provided
+  d$info$source_user <- switch(!is.null(d2_session),
```
jason-p-pickering, line 26: This might be safer? 

`any(class(d2_default_session) %in% "d2Session")`

R/createKeychainInfo.R
```diff
@@ -143,13 +143,20 @@ createKeychainInfo <- function(submission_path = NULL,
...
   # Determine country uids ####
   if (is.null(d$info$country_uids)) {
     d$info$country_uids <-
       unPackCountryUIDs(submission_path = d$keychain$submission_path,
                         tool = d$info$tool)
   }
 
+  ## Determine additional Organisation Unit information and save
+  ## under `d$info$organisation_unit` for use in validating mechanisms
+  d$info$organisation_unit <- getOUFromCountryUIDs(d$info$country_uids)
```
jason-p-pickering, line 158: See my comment above.

jason-p-pickering (changes requested):
Left a few comments.

DESCRIPTION
cnemarich: Turns out that `validateSchema` isn't called anywhere in our package, including the `data-raw` files, and also isn't called in the app, so I think this is fine to leave as suggested for now.

R/checkMechanisms.R
cnemarich:
```suggestion
    d$info$has_error <- TRUE
```

R/createKeychainInfo.R
cnemarich: I just tried this in a fresh R session and got this error:
```
Error in class(d2_default_session) %in% "d2Session" : 
  object 'd2_default_session' not found
```

If we want to require being logged into DATIM for this function, I think that will work, but if the user isn't logged in I think this will create an error.

R/checkPSNUData.R
cnemarich: It actually looks like this `datasets_uid` object hasn't been needed since April of last year when it was removed from the `datimvalidation::validateData` function call:
https://github.com/pepfar-datim/datapackr-app/commit/00cee25526612db5dd6f74ed41d87050ab59fbe1#diff-60bef5af43a1bd541299ce9bceeb226afb23deaa5e637b145c85e05ffd552b0cL84

Maybe we can just remove lines 56-66 if it's not called anywhere anymore.

R/checkPSNUData.R
cnemarich: That seems like a good idea. We would just have to document it, which I'm not sure quite how to do with a nested object like that, and also change it to be saved as an `.rda` file.

R/checkMechanisms.R
cnemarich: Yes, this wasn't edited when I moved that into the `createKeychainInfo` function. So this could be removed.

```suggestion

```

R/checkMechanisms.R
```diff
@@ -0,0 +1,57 @@
...
+  mechs_data <- unique(d$datim$MER$attributeOptionCombo)
+
+  period_info <- datimvalidation::getPeriodFromISO(paste0(d$info$cop_year, "Oct"))
+
+  operating_unit <- getOperatingUnitFromCountryUIDs(d$info$country_uids)
+
+  mechs_datim <- datapackr::getMechanismView(d2_session = d2_session,
+                                             update_stale_cache = TRUE,
+                                             cached_mechs_path = cached_mechs_path) %>%
+    dplyr::filter(ou == operating_unit$ou) %>%
```
cnemarich, line 25: See comment above. This would reference metadata in `d` object.
```suggestion
    dplyr::filter(ou == d$info$operating_unit$ou) %>%
```

R/createKeychainInfo.R
cnemarich: I think I've addressed your comment above with the suggested changes, but let me know if I misunderstood.

JordanBalesBAO (commented):
Code reviewed and everything looks fine. Moving to functional testing.

R/checkPSNUData.R
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+    vr_data <- datimvalidation::prepDataForValidation(vr_data) %>%
+      dplyr::select(-dataElement,-period,-categoryOptionCombo) %>%
+      dplyr::group_by(orgUnit,attributeOptionCombo) %>%
+      tidyr::nest()
+
+
+    #Evaluate the indicators in parallel if possible
+    if ("parallel" %in% rownames(installed.packages()) == TRUE) {
```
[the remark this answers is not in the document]
cnemarich: The `renv.lock` file would handle whether this package was installed in the first place, but can't handle whether to conditionally execute this code, which is why this is needed here.

JordanBalesBAO (approved):
Left more details on the Jira Ticket. Went through code changes with create detail, and tested front end functionality via the app. Looks good to me.

flopez-bao (approved):
Looked through the files and nothing stood out to me that looked out of place. I tested India and Indonesia 22 as well as Sudan without issue.

PR #382, opened 2022-02-15
Add support for COP22 data in getCOPDataFromDATIM

sam-bao:
This PR simply allows COP22/FY23 data to be returned by getCOPDataFrom DATIM.

No special handling is required for COP22 data in the function, but a note was added because it is likely special handling will needed to be added for COP23.


jason-p-pickering:
Closing this one as it was better to merge into DP-504 via https://github.com/pepfar-datim/datapackr/pull/383/

PR #393, opened 2022-03-01
Make checkColStructure more robust

jason-p-pickering:
- Handles situations a bit better when the structure of the PSNUxIM tab has been modified. It seems we more or less allow columns to be deleted (with a warning) but if columns were deleted, we got an unhanded error. 
- The changes allow for more flexibility to process even mangled PSNUxIM tabs, providing better user feedback as to the potential source of the problem. 

### Related Issues
- DP-548
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.

### DIFF
--- a/R/checkColStructure.R
+++ b/R/checkColStructure.R
@@ -16,6 +16,8 @@ checkColStructure <- function(d, sheet) {
     data <- d$data$extract
   }
 
+  to_keep_regex <- "\\d{4,}_(DSD|TA)|^$|^(Not PEPFAR)$"
+
   submission_cols <- names(data) %>%
     tibble::enframe(name = NULL) %>%
     dplyr::select(indicator_code = value) %>%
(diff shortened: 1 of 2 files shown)


R/checkColStructure.R
```diff
@@ -49,14 +56,14 @@ checkColStructure <- function(d, sheet) {
 
     warning_msg <-
       paste0(
-        "WARNING! In tab ",
+        "ERROR! In tab ",
```
jason-p-pickering, line 59: I elevated this to an error, but not really sure. Perhaps we should warn and proceed? Otherwise, we need to decide once and for all how to handle PSNUxIM tabs which do not exactly match the template structure.

flopez-bao: I think this should be an error. It will force people to fix issues before they reach the PSNUxIM tab.

flopez-bao: I think this should be an error. It will force people to fix issues before they reach the PSNUxIM tab.

PR #395, opened 2022-03-03
Missing combos fix

jason-p-pickering:

- Fixes issues with indicators (other than TX_CURR) which have fine ages in the main tabs. The app was incorrectly flagging missing PSNUxIM combos. 
-


sam-bao (approved):
Fausto has reviewed and tested, so approving.

PR #398, opened 2022-03-04
Dp 523

flopez-bao:
Changed WARNING and ERROR message metadata to match the message.
Testing currently shows warnings and errors now sorted and colored properly.
Tested each error type:
-Missing Sheets
--> Zambia datapack deleted a sheet
-Missing Prioritizations
--> Data Pack Zambia with modified blank prioritizations.
-Non-numeric Values
--> Data Pack_Zambia_2022_02_15_0824.xlsx
-Decimal Values
--> Data Pack_Zambia_2022_02_15_0824.xlsx
-Fatally Missing Columns
--> Indonesia ROP22 datapack_Feb 27_8PMv2.xlsx
-Imbalanced Distribution
--> SA_datapack_20220304_142710_new psnuxim.xlsx
-Unallocated IMs
--> SA_datapack_20220304_142710_new psnuxim.xlsx

### Related Issues
- DP-523

### DIFF
--- a/R/checkStructure.R
+++ b/R/checkStructure.R
@@ -30,7 +30,8 @@ checkStructure <- function(d) {
     warning_msg <-
       paste0(
         "ERROR! MISSING SHEETS: Please ensure no original sheets have",
-        " been deleted or renamed in your Data Pack. No targets will be imported for any tabs that have been deleted -> \n  * ",
+        " been deleted or renamed in your Data Pack. No targets will be",
+        " imported for any tabs that have been deleted -> \n  * ",
         paste0(d$tests$missing_sheets$sheet_name, collapse = "\n  * "),
         "\n")
     d$info$messages <- appendMessage(d$info$messages, warning_msg, "ERROR")
--- a/R/checkStructure.R
+++ b/R/checkStructure.R
@@ -33,7 +33,7 @@ checkStructure <- function(d) {
         " been deleted or renamed in your Data Pack. -> \n  * ",
         paste0(d$tests$missing_sheets$sheet_name, collapse = "\n  * "),
         "\n")
-    d$info$messages <- appendMessage(d$info$messages, warning_msg, "ERROR")
+    d$info$messages <- appendMessage(d$info$messages, warning_msg, "WARNING")
   }
 
   d
--- a/R/checkStructure.R
+++ b/R/checkStructure.R
@@ -29,11 +29,11 @@ checkStructure <- function(d) {
   if (any(NROW(d$tests$missing_sheets))) {
     warning_msg <-
       paste0(
-        "WARNING! MISSING SHEETS: Please ensure no original sheets have",
-        " been deleted or renamed in your Data Pack. -> \n  * ",
+        "ERROR! MISSING SHEETS: Please ensure no original sheets have",
+        " been deleted or renamed in your Data Pack. No targets will be imported for any tabs that have been deleted -> \n  * ",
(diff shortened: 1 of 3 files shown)


R/unPackSNUxIM.R
```diff
@@ -730,7 +730,7 @@ unPackSNUxIM <- function(d) {
         paste(imbalanced_distribution_inds, collapse = "\n\t* "),
         "\n")
 
-    d$info$messages <- appendMessage(d$info$messages, warning_msg, "WARNING")
+    d$info$messages <- appendMessage(d$info$messages, warning_msg, "ERROR")
```
jason-p-pickering, line 733: Not sure this should be an ERROR since we can recover from it. The values will be rounded. Need guidance from @jacksonsj here.

R/unPackSheet.R
```diff
@@ -184,7 +184,8 @@ unPackDataPackSheet <- function(d, sheet) {
     d$data$extract %<>%
       dplyr::filter(!stringr::str_detect(PSNU, "^_Military"),
                     # Excuse valid NA Prioritizations
-                    value != "NA")
+                    #value != "NA"
```
jason-p-pickering, line 187: Why is this commented out?

jason-p-pickering (changes requested):

R/unPackSNUxIM.R
```diff
@@ -539,7 +539,7 @@ unPackSNUxIM <- function(d) {
         paste(unique(d$tests$decimals$mechCode_supportType), collapse = "\n\t* "),
         "\n")
 
-    d$info$messages <- appendMessage(d$info$messages, warning_msg, "WARNING")
+    d$info$messages <- appendMessage(d$info$messages, warning_msg, "ERROR")
```
cnemarich, line 542: Change to WARNING not ERROR

R/unPackSNUxIM.R
datafaust: confirmed with @jacksonsj during standup, this will be changed to WARNING, text will be changed to match.

R/checkStructure.R
```diff
@@ -33,7 +33,7 @@ checkStructure <- function(d) {
         " been deleted or renamed in your Data Pack. -> \n  * ",
```
cnemarich, line 33: Should update this text to say that no targets will be imported for any tabs that have been deleted.

R/checkStructure.R
```diff
@@ -33,7 +33,7 @@ checkStructure <- function(d) {
...
         paste0(d$tests$missing_sheets$sheet_name, collapse = "\n  * "),
         "\n")
-    d$info$messages <- appendMessage(d$info$messages, warning_msg, "ERROR")
+    d$info$messages <- appendMessage(d$info$messages, warning_msg, "WARNING")
   }
 
   d
```
cnemarich, line 39: Change this to be an ERROR

R/unPackSheet.R
datafaust: So I set this back to uncommented as it was in order to stay with our concept of one set of issues for each PR. But I think there may be an issue here; R is interpreting NA values as NA not "NA" so it ends up filtering out any blank prioritizations. This is at least what I am getting when I test this Zambia datapack with blank prioritizations -- if I am correct I suggest this be a different ticket. I will post more on the dev channel with images of my tests.

cnemarich (approved):
These changes look good and fit the ticket requirements.

PR #438, opened 2022-04-01
Add additional comparison functions

jason-p-pickering:

- Rework the dataPackVsDATIM function to make use of the analytics data which is prepared as part of the memo generation

### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


JordanBalesBAO:
@jason-p-pickering I see this is still in draft status. Do we need to try and get this in or is it obe? It relates to R/generateComparisonPSNUxIM.R

jason-p-pickering:
Suggest we close this. The point of this PR was to create two different types of comparisons...
1) Used in the automation scripts to generate import files (updates and deletes)
2) In the app, we need a slightly different format which is more human readable. I think the current version of the app already takes care of this. 

So, I suggest we just close this without merging it, and can revisit if needed.

PR #442, opened 2022-04-05
Remove remaining NAT/SUBNAT validation rules

jason-p-pickering:

- There were a number of validation rules related to NAT and SUBNAT data elements in the validation rule JSON file. 
- All of these rules are not relevant since to the DataPack process since they are rules which define relationships between different classes of disaggregations. They create false positives in the validation rules report. 
- TODO: It is still not clear at this point whether we need to import the coarser disaggregations from the finer age/sex disaggs.

### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


data-raw/cop22_validation_rules.json
```diff
@@ -270,21 +165,6 @@
                 "missingValueStrategy": "NEVER_SKIP"
             }
         },
-        {
-            "name": "PMTCT_STAT_SUBNAT (N, SUBNAT, Age/Sex) TARGET <= PMTCT_STAT_SUBNAT (D, SUBNAT, Age/Sex) TARGET",
```
cnemarich, line 274: This rule appears to be the only one not comparing different levels of disaggregation. Should it be kept?

cnemarich:
@jason-p-pickering I agree with all other changes with the possible exception of the one highlighted.

data-raw/cop22_validation_rules.json
jason-p-pickering: Fixed this, so will merge now.

PR #458, opened 2022-05-11
Adds `apply_excel_limits` function

cnemarich:
* Adds a new function `apply_excel_limits` which:
   * apply limits to rowcounts in written Excel files
   * apply a 3000 item limit to nested lists

### Related Issues
- DP-558

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


cnemarich:
Needs linting and a test written.

jason-p-pickering:
@cnemarich this was implemented in the app actually [here ](https://github.com/pepfar-datim/datapackr-app/blob/c18b7bc6b1d77fa1ae72b99de4baec202cc54487/server.R#L662)

Do you want to re-implement this as a generic method in datapackr?

cnemarich:
@jason-p-pickering, yes since @flopez-bao has written the code to create this as a generic function already, I think it's worth writing a test or two and getting this into the package. Seems potentially very useful so long as we continue to work with Excel.

jacksonsj (approved):

jacksonsj:
Kudos on the use of lapply & map. Remind me where you plan to use this function?

flopez-bao:
If I remember correctly this was a way to apply the standard excel limit on characters in a cell which had been an issue in the vr rules. I think we found another way around this, but might be useful to keep this one around.

jason-p-pickering:
@jacksonsj the idea here is that anytime we export some object to Excel there should be

-  less than 32,767 characters per cell
- less than 16,384 columns
- less than 1,048,576 rows
- no nested lists in any columns

I took a slightly different approach in the app [here](https://github.com/pepfar-datim/datapackr-app/blob/17558e6fb03d90c6d7030d2903fda16ab237d608/server.R#L674) which I think does all of this. 

This is the only place we have run into problems that I know of when exporting to Excel, but I think the idea with this function is to basically resize/reshape a data frame into something which can fit inside Excel. 

Agree with @cnemarich . We should write some tests here.

JordanBalesBAO:
It would seem that this is taken care of in the app already. @jason-p-pickering @flopez-bao any pushback on closing this PR out?

jason-p-pickering:
I suggest we close this and revisit if there is a need in places other than the Datapack app.

PR #459, opened 2022-05-11
Add top line SUBNAT values to export

cnemarich:
### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.

### DIFF
--- a/R/checkAnalytics.R
+++ b/R/checkAnalytics.R
@@ -1,6 +1,4 @@
 
-
-
 HTS_POS_Modalities <- function(cop_year) {
     #TODO: This function needs a parameter based on COP year.
     #More work further down, so I am not going to fix it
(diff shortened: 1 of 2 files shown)


jason-p-pickering:
I think we need to consult with the DATIM and downstream teams to confirm that there is actually a need for this data. Now that we have already imported most (if not all ) DataPacks, we will probably need to wait until everything is imported. @cnemarich do you want to check with DATIM/PDAP/Program to clarify that we need these data elements before we do any more work here?

jason-p-pickering:
Assigning to you @cnemarich for requirements clarification. If we need to do this now, we are going to need to adopt a different approach by calculating this data from the data which has already been imported. If we do not need to do it, then I think we can close this ticket.

jason-p-pickering:
@christian-domaas I am beginning to think this is OBE. Do you or @jacksonsj have any further clarification whether these numbers needed to get into DATIM?

JordanBalesBAO:
@christian-domaas Any thoughts or can we close this out OBE?

jason-p-pickering:
Lets close it and revisit if there really is a need for this elsewhere than the Datapack app.

PR #464, opened 2022-05-12
Clear warnings and notes from build checks

cnemarich:
- Adds missing documentation to functions
- Adds missing package declarations with functions
- Updates `datapackr.R` with additional global variables
- Updates `.Rbuildignore` to add `mechs.rds`
- Removes `R6` as a package dependency since it is not called directly

### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/utilities.R
```diff
@@ -663,7 +663,7 @@ formatSetStrings <- function(vec) {
   }
 
   vec <- sort(vec)
-  groups <- cumsum(c(0, diff(vec) > 1))
+  groups <- cumsum(c(0, base::diff(vec) > 1))
```
jason-p-pickering, line 666: Seems odd that we would need to call base explicitly. Are you sure about this?

cnemarich: Yeah, removed this. Isn't necessary.

PR #469, opened 2022-05-17
Adds additional unit testing

jason-p-pickering:
### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/pullFullCodeList.R
```diff
@@ -25,43 +25,17 @@ pullFullCodeList <- function(FY = getCurrentCOPYear() + 1,
   datasets_provided <- !is.null(datasets)
 
   if (!datasets_provided) {
-    datasets <- character(0)
```
jason-p-pickering, line 28: @jacksonsj this was the simplification I had made to pullFullCodeList.

R/utilities.R
```diff
@@ -216,114 +216,73 @@ addcols <- function(data, cnames, type = "character") {
...
-    }
-    if ("impatt" %in% type) {
-      datasets <- c(datasets,
-                    "jxnjnBAb1VD") # Planning Attributes: COP Prioritization SNU FY2021
-    }
-  } else if (fiscal_year == "2020") {
-    if ("mer_targets" %in% type) {
-      datasets <- c(datasets,
-                    "sBv1dj90IX6", # MER Targets: Facility Based FY2020
+getDatasetUids <-  function(fiscal_year, type) {
```
jason-p-pickering, line 219: @jacksonsj here is the simplification I made for getDatasetUids. The behavior is slightly different though. Before FY2023 would return an error when no "type" was supplied. I am not sure this was intentional, but we can replicate this behavior if you want. The other possibility is to simply return a NULL, and let the consumer function deal with it.  Let me know what you think.

PR #480, opened 2022-05-25
DP-725: Create wrapper function `packTool`

cnemarich:
Creates new function `packTool` to serve as a wrapper for `packDataPack` and `packOPUDataPack` and contains common functionality between the two:
* Updates `check_params` to add check for `output_folder` that checks if folder path is provided and valid.
* Refactors `packDataPack` and `packOPUDataPack` to work with `packTool`. Now each function only takes in `d` and `d2_session` as parameters. Common functionality has been removed and moved into `packTool`. Each function now checks if appropriate model filepath has been provided and then writes tabs into pre-generated file.
* Updates `createDataPack` function to add `sane_name` and `operating_unit` to newly created `d` objects.
* Deprecates `compareTemplateToSchema` after removing its only remaining usage in `packDataPack`.

### Related Issues
- DP-725

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


data-raw/COP21OPU_Data_Pack_generation_script.R
```diff
@@ -1,72 +1,27 @@
 library(datapackr)
-library(magrittr)
 
 # Point to DATIM login secrets ####
 secrets <- Sys.getenv("SECRETS_FOLDER") %>% paste0(., "datim.json")
 datimutils::loginToDATIM(secrets)
 
 output_folder <- Sys.getenv("OUTPUT_FOLDER") %>% paste0(., "COP21 OPUs/")
 
-# batch <- tibble::tribble(
```
JordanBalesBAO, line 10: This comment might be useful to leave in.

data-raw/COP22_Data_Pack_generation_script.R
```diff
@@ -1,73 +1,32 @@
...
-#       "Namibia",
-#       "Zambia",
-#       "South Sudan",
-#       "Malawi"
-#     )
-#   )
+  dplyr::filter(datapack_name %in% generation_list)
 
-# For individual testing ####
-# pick <- datapackr::COP21_datapacks_countries %>%
```
JordanBalesBAO, line 36: Same as above, we might want to leave these in here.

JordanBalesBAO:
Looks good to me, just might want to consider leaving in some legacy comments.

data-raw/COP21OPU_Data_Pack_generation_script.R
cnemarich: I think this is now redundant to the `generation_list` which filters by OU name, rather than the previous system of filtering by row number.

data-raw/COP22_Data_Pack_generation_script.R
cnemarich: Same as above.

PR #506, opened 2022-06-02
Dp 360

cnemarich:
### Related Issues
- DP-360

## Reviewer:
- [x] Clean linting.
- [x] Well commented.
- [x] Build check passes.


R/utilities.R
```diff
@@ -674,37 +674,53 @@ formatSetStrings <- function(vec) {
...
+is_uidish <- function(string, ish = FALSE) {
+  if (!ish) {
+    stringr::str_detect(string, "^[[:alpha:]][[:alnum:]]{10}$")
+  } else {
+    stringr::str_detect(string, "[[:alpha:]][[:alnum:]]{10}")
+  }
 
+}
 
-extractWorkbook <- function(d) {
```
jason-p-pickering, line 688: Any reason this was deleted? Leading to test failures.

R/unPackDataPack.R
```diff
@@ -35,26 +35,30 @@ unPackDataPack <- function(d,
...
   # Separate Data Sets ####
   interactive_print("Separating datasets...")
-  d <- separateDataSets(d)
+  datasets <- separateDataSets(data = d$data$targets,
+                               cop_year = d$info$cop_year,
+                               tool = d$info$tool)
+
+  d$data$MER <- datasets$MER
+  d$data$SUBNAT_IMPATT <- datasets$SUBNAT_IMPATT
+  d$data <- within(d$data, rm("targets"))
```
cnemarich, line 61: Is there a particular reason why these `d$data` assignments were moved outside of the `separateDataSets` function so this process is no longer self-contained?

R/createKeychainInfo.R
```diff
@@ -56,50 +60,77 @@ createKeychainInfo <- function(submission_path = NULL,
     dplyr::mutate(
       cop_year = stringr::str_extract(home_cell, "COP\\d{2}"),
       cop_year = as.numeric(stringr::str_replace(cop_year, "COP", "20")),
-      type = stringr::str_extract(home_cell, "OPU Data Pack|Data Pack")
-    )
+      tool = stringr::str_extract(home_cell, "OPU Data Pack|Data Pack"))
+
+  # Is this even a DataPack tool? ----
+  if (!tool_metadata$cop_year[1] %in% c(2018:2030)
```
jason-p-pickering, line 66: Is there a specific reason that we do not use supported cop years here?

R/unPackDataPack.R
jacksonsj: Yes. Before, the input was d and it was hardcoded to look for a certain dataframe under d$data. But the basic functionality is really a lot bigger: take data and separate it out by DATIM dataset (e.g., MER, SUBNAT, etc.). Making the data itself the input instead of d makes this function a lot more useful. But then if you're splitting one to many, the output has to be a list object. Hence...

R/utilities.R
jacksonsj: Replaced with loadDataPack. Tests now updated

jacksonsj:
From @jason-p-pickering:

Tried to load the Rwanda datapack and saw this in the console. Its not clear to me 1) Why we are getting this warning at all 2) Why we are getting it multiple times.
image

jacksonsj:
From @jason-p-pickering:

image

from the Nigeria datapack

jacksonsj:
From @jason-p-pickering:

One other thing I have noticed, is that the test are running quite a bit slower. I think this is because each time we load a datapack in the tests, we run checkSheetData. I saw this comment, but not sure I agree with it:

```
  # Implementing this here instead of in unPackDataPack or unPackTool because
  # while you may want to checkSheetData without running unPackSheets, you should
  # should never unPackSheets without running checkSheetData
  d <- checkSheetData(d, sheets = sheets)
```

i think we should be able to skip the checks (since they take time and are tested elsewhere) during the CI run. So, if we added an additional parameter to loadDataPack and all dependent functions called check_sheet_data = TRUE (but set it to false in the unit tests), I think we should be able to safely skip those checks when loading from one of our test datapacks?

jacksonsj:
From @jason-p-pickering:

Next problem...performance

i checked against master with the Rwanda datapack...

14 seconds in master, and 28 seconds with DP-360

my impression previously was that this branch was faster, but now the parsing time appears to have doubled....

jacksonsj:
@jason-p-pickering - I'm really not sure where to go next on the performance issues you raised. This approach is definitely slower, I think because it requires loading and munging each sheet for each test performed, rather than before when tests were integrated in `unPackSheet` and so loading/munging happened just once upfront.

Options:
- If we alter the check functions to return a list of just `test_results` & `msg`, like in the `checkAnalytics` family, instead of the full `d` object, we could use `purrr::map` to run checks simultaneously. As is though, because each check alters `d` and then the next check uses that as input, we have to run each test for each sheet sequentially, which rules out `purrr::map`.
- Reverse course and reintegrate checks into `unPackSheet` as it was to begin with.
- Alter the checks further to allow them to be run against multiple `sheets` at once, rather than only against a single `sheet` at a time. For some checks, this would mean iterating over each sheet somehow and appending messages and results in a way that aligns with the new `messageQueue` approach. For other checks, you could potentially combine all data together from all sheets and check all at once, then split messages by a `sheet_name` column or something like that.

Thoughts?

jason-p-pickering:
Are any tests dependent on other tests?

jacksonsj:
@jason-p-pickering - Could you take a look at the new approach here, in `checkSheetData`? I've converted the underlying check functions to return a list of just test results, msg, lvl, & has_error, and then checkSheetData compiles these using map functions into d.

tests/testthat/test-unPackingChecks.R
```diff
@@ -0,0 +1,121 @@
...
+  expect_true(any(grepl("ERROR! In tab Prioritization: INVALID PRIORITIZATIONS", d$info$messages$message)))
+  expect_true(any(grepl("ALTERED FORMULAS", d$info$messages$message)))
+  expect_true(any(grepl("ERROR! In tab OVC: INVALID DISAGGS", d$info$messages$message)))
+
+  expect_gt(nrow(d$tests$duplicate_rows), 0)
+  expect_gt(nrow(d$tests$missing_cols), 0)
+  expect_gt(nrow(d$tests$duplicate_columns), 0)
+  expect_gt(nrow(d$tests$columns_out_of_order), 0)
+  expect_gt(nrow(d$tests$non_numeric), 0)
+  expect_equal(nrow(d$tests$negative_values), 6L)
```
jason-p-pickering, line 113: I was explicit about the number of rows which should be detected. Should we add this to all of the tests?

jason-p-pickering (changes requested):
Looks good @jacksonsj.Could you reply to my questions? Otherwise, I think this is solid work.

R/unPackingChecks.R
```diff
@@ -0,0 +1,1225 @@
...
+  }
+
+  # Get data
+  data <- unPackDataPackSheet(d,
+                              sheet = sheet,
+                              clean_orgs = TRUE,
+                              clean_disaggs = TRUE,
+                              clean_values = FALSE)
+
+  # keep_cols <- d$info$schema %>%
```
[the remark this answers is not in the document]
jacksonsj: Done

R/unPackSNUxIM.R
```diff
@@ -380,9 +380,9 @@ unPackSNUxIM <- function(d) {
 
   # TEST: Missing key metadata; Error; Drop ####
   # TODO: Make compatible for OPUs
-  if (d$info$tool == "Data Pack") {
-    d <- checkMissingMetadata(d, sheet)
-  }
+  # if (d$info$tool == "Data Pack") {
```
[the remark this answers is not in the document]
jacksonsj: Removed.

R/unPackingChecks.R
```diff
@@ -0,0 +1,1225 @@
...
+    # Decimal values ----
+    d <- checkDecimalValues(d, sheet)
+
+    # Check invalid org units ----
+    d <- checkInvalidOrgUnits(d, sheet)
+
+    # Check for invalid prioritizations ----
+    d <- checkInvalidPrioritizations(d, sheet)
+
+    # TEST AGYW Tab for missing DSNUs ####
```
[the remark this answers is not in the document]
jacksonsj: Need this one for future PR, but noted better in code.

R/unPackingChecks.R
```diff
@@ -0,0 +1,1225 @@
...
+}
+
+
+
+#' @export
+#' @rdname unPackDataChecks
+checkDisaggs <- function(d, sheet, quiet = TRUE) {
+
+  if (sheet %in% c("SNU x IM", "PSNUxIM")) {
+    stop("Sorry! Can't check the PSNUxIM tab with this function.")
```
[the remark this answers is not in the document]
jacksonsj: Will be once DP-472 is complete.

R/unPackingChecks.R
```diff
@@ -0,0 +1,1225 @@
...
+
+  rm(params, p)
+
+  # TODO: Apply method used in checkAnalytics line 631
+
+  sheets <- sheets[!sheets %in% c("KP Validation")]
+
+  for (sheet in sheets) {
+    # Col Structure ----
+    d <- checkColumnStructure(d, sheet)
```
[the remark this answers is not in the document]
jacksonsj: This code is outdated — was changed a week ago or so. This `for` loop isn't there anymore. Using this approach now:

```R
# Apply the list of check functions ----
  funs <- list(
    duplicate_rows = checkDupeRows,
    missing_cols = checkMissingCols,
    duplicate_columns = checkDupeCols,
    columns_out_of_order = checkOutOfOrderCols,
    non_numeric = checkNonNumeric,
    negative_values  = checkNegativeValues,
    decimal_values = checkDecimalValues,
    invalid_orgunits = checkInvalidOrgUnits,
    invalid_prioritizations = checkInvalidPrioritizations,
    altered_formulas = checkFormulas,
    defunct_disaggs = checkDisaggs
  )

  data_checks <-  purrr::map(funs, purrr::exec, sheets, d)

  d$tests <-
    append(d$tests,
           purrr::map(data_checks, ~ purrr::pluck(.x, "result"))) %>%
    purrr::discard(is.null)
```

tests/testthat/test-unPackingChecks.R
jacksonsj: Done

jacksonsj:
Just a note that right now there is a warning in tests about missing parent prioritizations. Right now, the COP21 test sheet we use purposefully assigns a `29` as prioritization for one of the DSNUs — Berea. But `imputePrioritization` interprets this as a missing prioritization and flags as a warning. Also, this function checks for the list of expected prioritizations against the import file’s `orgUnit` column — if a PSNU gets left out for some reason across an entire Data Pack and ends up with no prioritization, `imputePrioritization` wouldn’t pick this up. I think we need to somehow better link `imputePrioritization` and `checkInvalidPrioritizations` (which detects missing and invalid prioritizations)

PR #519, opened 2022-07-20
DP-607: Refactor `datapackr` to utilize `datimutils::getDataValueSets`

cnemarich:
- Deprecates `getDataValueSets` and any related tests
- Replaces usage of internal `getDataValueSets` function with `datimutils::getDataValueSets`

### Related Issues
- DP-607

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


JordanBalesBAO:
I need to unit test this again before we spend too much time testing

R/getCOPDataFromDATIM.R
```diff
@@ -66,12 +66,6 @@ getCOPDataFromDATIM <- function(country_uids,
     return(NULL)
   } else {
     datim_data %>%
-    dplyr::rename(
-      dataElement = data_element,
-      orgUnit = org_unit,
-      categoryOptionCombo = category_option_combo,
-      attributeOptionCombo = attribute_option_combo
-    ) %>%
```
cnemarich, line 74: Will removing the renaming here break things downstream?

JordanBalesBAO:
@cnemarich the renaming was rather odd the function would change the names, yet the down stream functions would change them back.  So no it shouldn't. 

I am going to wrap up functional testing today. The current function uses these paramaters, so shouldn't be too much left to test. 

image

jason-p-pickering:
It should not break things downstream, because the datimutils function supplies these names as is. The equivalent datapackr function supplied the snake cased names, which we then renamed again. Something may break, but if so, we need to find it, since the unit tests are passing.

R/datapackVsDatim.R
```diff
@@ -273,7 +273,7 @@ compareData_OpuDatapackVsDatim <-
     # get data from datim using dataValueSets
     # rename to standard names
     datim_data <-
-      getDataValueSets(parameters$key,
+      datimutils::getDataValueSets(parameters$key,
                        parameters$value,
                        d2_session = d2_session) %>%
       dplyr::rename(
```
JordanBalesBAO, line 279: @jason-p-pickering my local branch pretty much matched what you did, the only difference being I also took out this rename. Just wanted to make sure we should leave this in?

jason-p-pickering: I am not sure I understand what you mean? Which renaming?

JordanBalesBAO: Line 279 in datapackVSDatim.R, after running getDataValueSets we dplyr::rename the columns. Where as in getCOPDataFromDATIM.R we have removed the dplyr::rename.

JordanBalesBAO: So after reviewing the dplyr documentation and the Datim api I think we are good to remove 4/5 of the renames. I don't think we need tests for these functions as they are not part of the package's main purpose.

https://dplyr.tidyverse.org/reference/rename.html 

image

JordanBalesBAO (approved):

PR #524, opened 2022-08-01
Fix negative values test.

jason-p-pickering:
Speculative fix for tests for negative values.  I was not certain that the values in the `value` column were actually cast to numeric before the `value < 0` comparison, so used a string based regex comparison to detect values which start with `-`. This may explain the test differences in different enviornments.

### DIFF
--- a/R/unPackingChecks.R
+++ b/R/unPackingChecks.R
@@ -635,7 +635,7 @@ checkNegativeValues <- function(sheets, d, quiet = T) {
                                          clean_orgs = F,
                                          clean_disaggs = F,
                                          clean_values = F) %>% #Are the values numeric at this point?
-    dplyr::filter(stringr::str_detect(value,"^-"))
+    dplyr::filter(stringr::str_detect(value, "^-"))
 
   if (NROW(negative_values) > 0) {
 
--- a/R/unPackingChecks.R
+++ b/R/unPackingChecks.R
@@ -469,7 +469,8 @@ checkOutOfOrderCols <- function(sheets, d, quiet = TRUE) {
             lvl = NULL,
             has_error = FALSE)
 
-  out_of_order <- d$sheets[sheets] %>%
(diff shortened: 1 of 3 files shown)


jason-p-pickering:
So, the other possibility is that we convert to numeric first...then do the comparison, which the last commit does. However, would it be best to go ahead and do this numeric conversion only once? ....perhaps for a future optimization ticket.

R/unPackingChecks.R
```diff
@@ -469,7 +469,7 @@ checkOutOfOrderCols <- function(sheets, d, quiet = TRUE) {
             lvl = NULL,
             has_error = FALSE)
 
-  out_of_order <- d$sheets[sheets] %>%
+  out_of_order <- d$sheets[names(d$sheets) %in% sheets] %>%
```
jacksonsj, line 472: | expr | mean (ms) |
| --- | --- |
| `d$sheets[names(d$sheets) %in% sheets]` | 0.00500427 |
| `d$sheets %>% purrr::keep(names(.) %in% sheets)` |  0.01475524|

R/unPackingChecks.R
```diff
@@ -632,9 +632,13 @@ checkNegativeValues <- function(sheets, d, quiet = T) {
...
-                                         clean_orgs = F,
-                                         clean_disaggs = F,
-                                         clean_values = F) %>%
+                                         clean_orgs = FALSE,
+                                         clean_disaggs = FALSE,
+                                         clean_values = FALSE) %>%
+    #TODO: Keeping this consistent with checkDecimalValues
+    #Consider doing the numeric conversion once in unPackDataSheet
+    #instead of multiple times in these checks.
+    dplyr::mutate(value = suppressWarnings(as.numeric(value))) %>%
```
jacksonsj, line 641: Woops. Yeah, that would've helped. A lot. Thanks.

jacksonsj (approved):

PR #529, opened 2022-08-03
Fix imputePrioritizations

jacksonsj:
### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/adorn_import_file.R
```diff
@@ -65,76 +66,105 @@
...
-         add_dp_psnu() %>% #Found in getPSNUs.R
-         dplyr::select(ou, ou_id, country_name, country_uid, snu1, snu1_id,
-                       psnu, psnu_uid, dp_psnu, psnu_type, DREAMS)), #cols to keep
-      by = c("orgUnit" = "psnu_uid")) #Columns to join on
+  row_num <- NROW(psnu_import_file)
+
+  cop_year %<>% check_cop_year()
+
+  # Adorn orgunits ----
+  snus <- valid_PSNUs %>% # Comes from file data/valid_PSNUs.rda
```
jason-p-pickering, line 93: Can you move all of this handling outside of the function so that we can test it separately? It looks like you already want to deprecate imputePrioritizations anyway...why not just replace it with this new code so its more easily unit tested?

R/adorn_import_file.R
```diff
@@ -65,76 +66,105 @@
...
-    data %<>%
-      dplyr::left_join(prio, by = "orgUnit") %>% # Join data and prio
+    psnu_prioritizations %<>%
+      dplyr::select(orgUnit, value) %>%
+    # Remove anything not a PSNU & remove Mil
+      dplyr::filter(orgUnit %in% snus$psnu_uid[snus$psnu_type != "Military"],
+    # Remove invalid values
+                    value %in% prioritization_dict()$value) %>%
+    # Translate values into descriptions
+      dplyr::left_join(prioritization_dict() %>%
```
jason-p-pickering, line 143: So, I think with the left join, the assert statement is always going to work, but, the situation we want to avoid is that we end up with a Dreams SNU with no prioritization, which seems like it could happen with a left join. See my comment above regarding unit testing.

jason-p-pickering (changes requested):
Needs more tests here I think for the new logic before merging.

jason-p-pickering:
I think this looks good, so I am going to go ahead and merge this into dev.

PR #532, opened 2022-08-10
DP-609: Retire internal API utility functions

cnemarich:
### Summary of Proposed Changes
Removes `api_version`, `api_get`, `api_filter`, `api_fields`, and `api_call` as all of their usages have been replaced by `datimutils` functions.

### Related Issues
- DP-609.

## Reviewer:
- [ ] Tests added/updated & passing.
- [ ] Clean linting.
- [ ] Related issue ticket in Jira/GitHub.
- [ ] Documentation added/updated.
- [ ] Code conforms to style guidelines.
- [ ] Well commented.
- [ ] Updates reflected in NEWS.md.
- [ ] Build check passes.

PR #523, opened 2022-07-27
DP-765: COP21/22 OPU tool generation & processing

cnemarich:
### Summary of Proposed Changes
- Implements the ability to pack and unpack COP22 OPU tools

### Related Issues
- DP-765

## Reviewer:
- [ ] Tests added/updated & passing.
- [ ] Clean linting.
- [ ] Related issue ticket in Jira/GitHub.
- [ ] Documentation added/updated.
- [ ] Code conforms to style guidelines.
- [ ] Well commented.
- [ ] Updates reflected in NEWS.md.
- [ ] Build check passes.

PR #537, opened 2022-09-20
DP 768: Convert valid_PSNUs to valid_OrgUnits and update format and columns

jacksonsj:
### Related Issues
- DP-768

## Reviewer:
- [x] Clean linting.
- [x] Well commented.
- [x] Build check passes.


jacksonsj:
@jason-p-pickering - This has two failing tests that seem unrelated to anything changed in this PR. Seem to be issues in `dev` too maybe?

```
══ Failed tests ═══════════════════════════════════════════════════════���════════
── Error (test-get-code-list.R:6:5): We can get a list of PSNUs from DATIM ─────
Error in `UseMethod("status_code")`: no applicable method for 'status_code' applied to an object of class "try-error"
Backtrace:
    ▆
 1. └─datapackr::getDataPackOrgUnits(...) at test-get-code-list.R:6:4
 2.   └─datimutils::getMetadata(...)
 3.     └─datimutils:::api_get(...)
 4.       └─httr::status_code(resp)
── Failure (test-prepareMemoData.R:93:5): We can create Datapack memo data ─────
all(names(d$memo$datapack$by_prio[3:prio_cols_end]) %in% prioritization_dict()$name) is not TRUE

`actual`:   FALSE
`expected`: TRUE
```

The first has something to do with this line in `datimutils`: https://github.com/pepfar-datim/datimutils/blob/master/R/callDATIMapi.R#L100. Maybe this is related to the use of the `training` instance in our tests? Not sure.

The second one is because in one of our test sheets, we have some org units without prioritizations, leading to `NAs`, which don't have a match in `prioritization_dict()`.

If you are able to help resolve, that would be great.

jacksonsj:
@JordanBalesBAO - If you have a second, would love your eyes on this too, particularly around the datimutils test failure.

JordanBalesBAO:
@jacksonsj Spent some time looking at your branch. Not sure what is occurring datimutils wise. 

- I was able to manually run the org unit functions that are failing testing. 
- I used training.datim as well as datim
- The api responses new vs old look the exact same
- The only difference I found was the test should be failing with the names check as it leaves out the "lastUpdated" column
- My guess is the try-catch we added to datimutils a month back isn't playing nice with the mock_api packages

JordanBalesBAO:
Fausto and I took a second look together, and we both agree the api mock package might not be able to replicate the response code that the try catch tests for

jason-p-pickering:
@jacksonsj I updated the renv.lock file to use the latest version of datimutils, and I was able to at least execute the tests. At least they are failing now, but I think you should be able to proceed, either by fixing the code, or tests.

jason-p-pickering:
@jacksonsj I have fixed the unit tests as well as other problems in the memo preparation. 

I am a bit puzzled though about the valid_orgUnits object. I sort of thought the point of this PR was to fetch the organisation units from the server. But if that is the case, why do we still have this object in the package? I must be missing something....

jason-p-pickering:
Ah, right...I think the rest of this is in https://github.com/pepfar-datim/datapackr/pull/532. 

Honestly, I think it would be easier to continue here with @cnemarich work instead of trying to merge that branch. Do you agree  @jacksonsj ?

jacksonsj:
I'm looking at this now. Not ready to merge quite yet

jacksonsj:
@jason-p-pickering - One test still failing, related to `training` API snapshots of `organisationUnits`. Not sure how those were created or how to fix. Mind taking a quick look?

R/getPSNUs.R
```diff
@@ -1,122 +1,166 @@
...
+  organisationUnits <-
+    datimutils::getMetadata(
+      "organisationUnits",
+      api_filters,
+      fields = fields,
+      d2_session = d2_session)
+
+  # orgunits <- datimutils::getSqlView(sql_view_uid = ""), #Replace once DP-786 resolved
+
+  # Filter by country if needed ####
```
jason-p-pickering, line 79: @jacksonsj I am not sure we want to do the filtering after the request. This means that when we are trying to get orgunits for Lesotho, we have to request them for Global, which is very expensive operation. Are you sure you want to do the filtering client side?

jacksonsj: Filtering client side is, surprisingly, x times faster than filtering via API. If we could set up a SQL View, that would be x times faster even than that:

```
datimutils::loginToDATIM("~/.secrets/datim.json")
d2_session <- d2_default_session

country_uids <- c("qllxzIjjurr", #Lesotho,
                               "XtxUYCsDWrR", #Rwanda
                              "cDGPF739ZZr") #South Africa

api_filters <- "organisationUnitGroups.id:in:[AVy8gJXym2D,nwQbMeALRjL,mRRlkbZolDR]"
filters_with_country <- c(api_filters,
                                           paste0("ancestors.id:in:[",
                                                        paste(country_uids, collapse = ","),
                                                         "]"))

fields <-
    "id,name,lastUpdated,ancestors[id,name,organisationUnitGroups[id,name]],organisationUnitGroups[id,name]"

microbenchmark::microbenchmark(
  datimutils::getMetadata(
    "organisationUnits",
    api_filters,
    fields = fields,
    d2_session = d2_session),
  datimutils::getMetadata(
    "organisationUnits",
    filters_with_country,
    fields = fields,
    d2_session = d2_session),
  datimutils::getSqlView(sql_view_uid = "gsaaxFM8ZN0"),
  times = 5L,
  unit = "s")
  ```

image

Comment (sam-bao) [reply R/getPSNUs.R:79]: It is much faster to get psnus (and mil and dreams) using the getOrgUnitGroups endpoint

```
datimutils::getOrgUnitGroups("AVy8gJXym2D", 
fields = "organisationUnits[name,id]")
```

File: R/adorn_import_file.R [hunk 2]
```diff
@@ -7,64 +8,64 @@
...
+#' @inheritParams datapackr_params
 #'
-#'
-#'
+#' @return Tibble of orgunits mapped to orgunit name, PSNU name, & PSNU uid.
+getPSNUInfo <- function(snu_uids,
+                        d2_session = dynGet("d2_default_session",
+                                                  inherits = TRUE)) {
 
-getPriorizationSNU <- function(psnu_uids) {
```
jason-p-pickering, line 14: Hmm...These get added back in #523 @jacksonsj

R/adorn_import_file.R
```diff
@@ -161,16 +199,17 @@ adorn_import_file <- function(psnu_import_file,
   # Stack data_codes and data_ids on top of one another.
   psnu_import_file <- dplyr::bind_rows(data_codes, data_ids, data_default) %>% dplyr::distinct()
   # Utilizes row_num to ensure the join,filter,stack worked as expected
-  assertthat::are_equal(NROW(psnu_import_file), row_num)
```
jason-p-pickering, line 164: Is there a good reason we are commenting out this assertion?

R/adorn_import_file.R
```diff
@@ -161,16 +199,17 @@ adorn_import_file <- function(psnu_import_file,
...
+  #assertthat::are_equal(NROW(psnu_import_file), row_num)
 
   # Adorn dataElements & categoryOptionCombos ####
 
   map_des_cocs <- getMapDataPack_DATIM_DEs_COCs(cop_year) # Found in utilities.R
 
-  # TODO: Is this munging still required with the map being a function of fiscal year?
```
jason-p-pickering, line 170: Remove this code if not needed.

R/createAnalytics.R
```diff
@@ -33,6 +33,22 @@ createAnalytics <- function(d,
   pzns <- d$datim$prioritizations %>%
     dplyr::select(orgUnit, value)
 
+  if (d$info$unallocatedIMs || !d$info$has_psnuxim) {
```
jason-p-pickering, line 36: Hmm...where did this come from?

renv.lock
```diff
@@ -327,16 +327,17 @@
     },
     "datimutils": {
       "Package": "datimutils",
-      "Version": "0.5.1",
+      "Version": "0.6.0",
```
jason-p-pickering, line 330: Noting the change to upgrade datimtutils here...

jason-p-pickering (changes requested):
Please have look at my comments before merging @jacksonsj

renv.lock
jacksonsj: This must've been added by Chris. Should I move it back to `0.5.1`?

R/createAnalytics.R
jacksonsj: This is in there twice and both came from something already in `dev`

JordanBalesBAO:
:+1::skin-tone-2:

PR #560, opened 2022-10-21
Dp 765 sjj freshstart

jacksonsj:
## Developer: @jacksonsj @cnemarich @jason-p-pickering 

@jason-p-pickering - This represents what you have in PR #555 , but focused solely on the generation code. Nothing about testing creation/unpacking of OPUs, memo prep, or DP-795 right side formulas in PSNUxIM tools. I'll break those others off into their own PRs too. Trying to chunk this up smaller to make it more manageable to get through.

### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/packTool.R
```diff
@@ -40,46 +40,28 @@ packTool <- function(model_data_path = NULL,
     assign(p, purrr::pluck(params, p))
   }
 
-  # Set global numeric format ####
-  options("openxlsx.numFmt" = "#,##0")
+  rm(params, p)
```
jason-p-pickering, line 43: @jacksonsj did we mean to remove the global numeric format here?

jason-p-pickering (approved):
@jacksonsj see my comment about global number format. We can add this back in later if needed....

R/packTool.R
jacksonsj: We do it in `createDataPack`, which we call in `packTool` already, so this line was redundant: https://github.com/pepfar-datim/datapackr/blob/121bef675538774087b600279e4fc70ebb78a1e9/R/utilities.R#L389

PR #582, opened 2023-01-11
feat: Introduce function for getting a valid organisation units by COP Year

jason-p-pickering:

- Introduces `getValidOrgUnits(cop_year). This function should return a data frame of valid organisation units for a given COP year. 
- Multiple changes in `check_params` to take this into account
- Minor updates to how invalid prioritization are detected

### Related Issues
- DP-788
- DP-837

## Reviewer:
- [x] Clean linting.
- [x] Well commented.
- [x] Build check passes.


R/createKeychainInfo.R
```diff
@@ -1,3 +1,19 @@
+
+unPackHomeTabMetadata <- function(submission_path)  {
```
jacksonsj, line 2: Can we separate this into a separate PR and keep this one focused on the getValidOrgUnits change?

R/evaluateIndicators.R
```diff
@@ -77,28 +79,42 @@ evaluateIndicators <- function(combis, values, inds) {
...
 
   #Function to replace missing combis with zeros
   replaceExpressionsWithZeros <- function(x) {
     expression.pattern <- "#\\{[a-zA-Z][a-zA-Z0-9]{10}(\\.[a-zA-Z][a-zA-Z0-9]{10})?\\}"
     gsub(expression.pattern, "0", x)
   }
 
+
+  #Pare down to only valid numeric expressions at this point
+  expr_regexes <- list(plus  = "\\+",
```
jacksonsj, line 91: Can we separate this into a separate PR and keep this one focused on the getValidOrgUnits change?

R/fetchPrioritizationTable.R
```diff
@@ -34,18 +35,18 @@ fetchPrioritizationTable <- function(psnus, cop_year,
     })
   }
 
-  n_requests <- ceiling(nchar(paste(psnus$psnu_uid, sep = "", collapse = ";")) / 2048)
+  n_requests <- ceiling(nchar(paste(psnus, sep = "", collapse = ";")) / 2048)
```
jacksonsj, line 38: Not sure I understand how this change fits in.

R/packDataPack.R
```diff
@@ -10,17 +10,35 @@
...
                          d2_session = dynGet("d2_default_session",
                                              inherits = TRUE)) {
 
-  # Checks and reads in Data Pack Model File ####
-  stopifnot(
-    "Model data file could not be read!" = canReadFile(d$keychain$model_data_path),
-    "Model data is not correct file type! File must have .rds extension." =
-      tools::file_ext(d$keychain$model_data_path) == "rds"
-  )
+  # is packDataPack receiving a model path or a model structure?
```
jacksonsj, line 17: Why the need to allow either path or file to be sent? Can we separate this into a separate PR and keep this one focused on the getValidOrgUnits change?

R/unPackSNUxIM.R
```diff
@@ -41,51 +41,20 @@ checkHasPSNUxIM <- function(d) {
...
-#' @return d
+#' @return Modified d object with documented and missing SNUxIM combos
 #'
-unPackSNUxIM <- function(d) {
-
-  sheet <- "PSNUxIM"
-
-  header_row <- headerRow(tool = d$info$tool, cop_year = d$info$cop_year)
+#' @example
+extractSNUxIMCombos <- function(d) {
```
jacksonsj, line 51: This is a pretty significant reworking of this function. Can we separate this into a separate PR and keep this one focused on the getValidOrgUnits change?

jacksonsj (changes requested):

R/packDataPack.R
flopez-bao: I think the purpose of this addition to packDataPack is from test coverage that was added [here](https://github.com/pepfar-datim/datapackr/blob/dev/tests/testthat/test-packDataPack.R). The purpose was to be able to feed in model data as shown [here](https://github.com/pepfar-datim/datapackr/blob/dev/tests/testthat/test-packTool.R#L17).  This adds transparency in the case of us seeing what exactly is model_data. This doesn't change anything in the logic other than to provide the option of passing an R object as opposed to a reading in an actual rds.

R/fetchPrioritizationTable.R
jason-p-pickering: This is just a simplification. There was no need to pass the entire orgunit data frame to this function. All we need is a vector of PSNU UIDs.

R/packDataPack.R
jason-p-pickering: DP-788 was lacking some commits from both master and dev. I will make two separate pull requests for these, so that it becomes more obvious what has changed in this PR.

R/createKeychainInfo.R
jason-p-pickering: I needed this again in `R/unPackCountryUIDs.R`

flopez-bao (approved):

R/evaluateIndicators.R
jason-p-pickering: I think this should be more clear once we merge with dev/master

R/unPackSNUxIM.R
jason-p-pickering: again, this was a diff from dev/master

PR #588, opened 2023-01-19
Dp 875

JordanBalesBAO:
Already reviewed and in master as was a hot fix. I do see some additional diffs from what I saw on the master branch TX PLV(N), and just more than the d object update. @jason-p-pickering is this expected?


JordanBalesBAO:
Oh, I never put the hot fix for 5.1.1 back into dev is the cause of this. I believe.

JordanBalesBAO:
Confirmed the last hot fix was put back in dev, so not the issue.  Will leave for now, until can discuss with team.

PR #601, opened 2023-02-02
feat: Add data object of data set levels

jason-p-pickering:
### Related Issues
- DP-843
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


tests/testthat/test-dataset-levels.R
```diff
@@ -0,0 +1,20 @@
...
+  expect_named(dataset_levels, expected_columns)
+  expect_true(all(stringr::str_detect(dataset_levels$iso3, "^[A-Z]{3}$")))
+  expect_true(all(stringr::str_detect(dataset_levels$iso4, "^[A-Z]{3}$")))
+  expect_false(all(is.na(dataset_levels$country_name)))
+  expect_false(all(is.na(dataset_levels$country_level)))
+  expect_true(all(dataset_levels$country_level >= 3 & dataset_levels$country_level <= 4))
+  expect_true(all(dataset_levels$community_level <= 8))
+  expect_true(all(dataset_levels$facility_level <= 9))
+  expect_true(all(dataset_levels$prioritization <= 7))
+  expect_true(all(dataset_levels$cop_year >= 2021 & dataset_levels$cop_year <= getCurrentCOPYear()))
```
JordanBalesBAO, line 16: Should we hard code 2021? Has no functional barring since it is greater than equal to, but just wanted to check.

JordanBalesBAO (approved):
Functionally I see no problem with it. It solves the problems set forth in DP-843.  

- How often will this run? I presume similar to how mechs work where if it is stale it will fetch a new one only then?
- Why is cop23 pulling from cop test, or is that just temporary?

tests/testthat/test-dataset-levels.R
jason-p-pickering: Agree..was kind of arbitrary. I have changed this to assert that the years in the data set levels should be in `supportedCOPYears()` . We may need to change this later on though if we have a situation where we might need dataset levels for COP years which we do not really "support". Anyway, let's deal with that later if we need to.

jason-p-pickering:
This object will need to be updated before each COP, or if a new country has been added to the hierarchy.

PR #695, opened 2023-05-15
Consolidate comparison function

jason-p-pickering:

- Removes `compareData_OpuDatapackVsDatim`
-  Adds logic to determine if SUBNAT/IMPATT data exists. If this is a hybrid tool  (Datapack + PSNU) then the function will compare SUBNAT/IMPATT data as well. However, if this is a standalone PSNU tool (aka OPU Data Pack), then the SUBNAT/IMPATT data will be excluded from the comparison analysis. 

### Related Issues
- DP-726

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


flopez-bao (approved):
Code looks good to me, but on this comment, is getCOPDataFromDATIM returning null for the current datapack year but not the previous year?

jason-p-pickering:
Yeah, so I added this because we have it in getCOPDataFromDATIM. My thinking here is that normally, we want to compare all data streams, except in the case of when we are working with  full datapack  + PSNU which is an OPU. In this case, we should ignore SUBNAT/IMPATT data entirely, since even if it has changed, we do not update it during an OPU. After talking with @flopez-bao yesterday, I added back `compareData_OpuDatapackVsDatim` for backwards compatibility. All this function does is to call `compareData_DatapackVsDatim(d, d2_session, datastreams = "mer_targets")`. That was the reason for this new parameter, if that makes sense.

jason-p-pickering:
@JordanBalesBAO I have fixed the merge conflict now.

flopez-bao:
So I have noticed an error flagging in unpacking cop23 datapacks with the use of unPackTool and unPackToolSet, specifically errors regarding Blank ORG Units (ERROR! In tab AGYW, INVALID OR BLANK ORG UNITS). I have seen this behavior with the following datapacks:
1. [Cameroon](https://datim.zendesk.com/agent/tickets/53417)
2. [Benin](https://datim.zendesk.com/agent/tickets/54246)
3. [Botswana](https://datim.zendesk.com/agent/tickets/53387)
4. [Burkina Faso](https://datim.zendesk.com/agent/tickets/54245)
This does not occur in Master. Can someone confirm similar results?

jason-p-pickering:
> specifically errors regarding Blank ORG Units (ERROR! In tab AGYW, INVALID OR > BLANK ORG UNITS).

I was able to reproduce this. Looks like (at least for Cameroon) this is being caused by some data in the sheet itself. There is one row in the AGYW tab for Cameroon, and pretty sure this is why this issue is being flagged. I am not really sure how to handle this, as the entire point of the test is to flag invalid or blank orgunits, which it is doing, but in this case, the row itself is not blank, so that is why it is not being filtered. I suspect it it the same issue for the other DataPacks.  

Not sure @flopez-bao . Do you think this is a new bug or perhaps we fixed one?

Screenshot from 2023-06-21 07-21-00

flopez-bao:
@jason-p-pickering I can see the same rows in those other datapacks as you mentioned and I have identified the "culprit". I believe it is this [line](https://github.com/pepfar-datim/datapackr/blame/dev/R/unPackingChecks.R#L794). This is from [dp953](https://github.com/pepfar-datim/datapackr/commit/64b844d59676a5632f6fe291b650d45c8f65303e#diff-5b1e386ece9731d90127d677e80cae0779904d458011d11e25262dbc3b61387b). I'm assuming this was a fix on your end since the behavior we are seeing is expected right? AKA an invalid row? 

If so this doesn't stop the import process but still flags ERROR in validation so I think we either:
1. expect the country team to delete that row
2. ignore it
3. functionality to toggle to warning if the invalid row is something like this and not an actual value?

jason-p-pickering:
Hmm. It is not clear to me why that line was commented out but I am inclined to leave it as it is. The line is clearly not valid, and my feeling here is that it should be deleted. Having said that, the test says there are two lines, when I think there should only be one. Let me have a look at it before we merge, but do you agree that this line should be flagged and deleted? Thoughts @jacksonsj ?

flopez-bao:
I agree that it should be deleted so we should leave it as is. That's the only thing I wanted to flag on this branch, but ultimately it's coming from dev so otherwise this branch seems good to go.

jacksonsj:
That line was commented out here: https://github.com/pepfar-datim/datapackr/blame/df4318edc93ece57ffbff5a28f76d234cb76d360/R/unPackingChecks.R

We need to look into it further, but agreed it can be in a follow on PR.

PR #698, opened 2023-05-18
Release 6.3.2

JordanBalesBAO:
## Developer: Team

Release Branch Review


JordanBalesBAO:
If we want the other 2 pull requests related to: 

- Tests
- CompareOPU 

Then we need to point them here **BEFORE** spending time reviewing this. Im okay pushing them to the next release, but I will heed to the developers who spent time coding them.

JordanBalesBAO:
I will need some help on the news.md

JordanBalesBAO:
Ready for review

R/unPackSNUxIM.R
```diff
@@ -885,7 +885,14 @@ unPackSNUxIM <- function(d) {
   }
 
   d$data$SNUxIM %<>%
-    dplyr::mutate(value = round_trunc(value))
+    dplyr::mutate(value = round_trunc(value)) %>%
+    dplyr::filter(value > 0)
```
jason-p-pickering, line 889: @JordanBalesBAO @flopez-bao @jacksonsj I am not sure we want to filter out values greater than zero here. Are we sure that we will get zero valued dedupes?

jason-p-pickering (changes requested):
I think we need to review the line that filters values greater than zero....

JordanBalesBAO (commented):
Completed my second round of code review

flopez-bao (commented):
Aside from the zero filter comment, looked through the code and ran a few 23 datapacks through. Only thing to flag is since this includes the dev branch, it flags invalid org units as errors for the row in most of the datapacks that has an NA value, same conversation in [pr-695](https://github.com/pepfar-datim/datapackr/pull/695). 

Don't think this is a problem, I mention it because that's really the only thing holding up PR-695, so if we can agree on that, we can merge 695 into this release and should be good to go.

JordanBalesBAO:
Lets discuss at standup in the morning and get 695 merged if so

R/unPackSNUxIM.R
jacksonsj: Looks like this change was committed by @jason-p-pickering on April 26 here: https://github.com/pepfar-datim/datapackr/commit/bd5bad5693c0cc7b9f5a1e61e2d5935312f7a660, and then reverted as part of this PR, yes?

jacksonsj: Tracking this issue here: https://jira.pepfar.net/browse/DP-1112

I vote we address in a follow-on PR and move this release forward.

JordanBalesBAO:
> conversation

Issue being tracked here,

https://jira.pepfar.net/browse/DP-1113

JordanBalesBAO:
The two stated "issues" have been captured in tickets. 

1. In regards to the non zero's this hot fix was backed out a long time ago, and this Release branch should reflect such
2. In regards to the blank row, based upon conversation between Fausto, Jason, and I this is how we want the software to perform. 

@jason-p-pickering  keep us honest, is this a good justification?

R/unPackSNUxIM.R
jason-p-pickering: I have an entire branch that started out addressing this, but has since morphed into something bigger. I agree we should track this issue in another ticket.

jason-p-pickering:
> 2. In regards to the blank row, based upon conversation between Fausto, Jason, and I this is how we want the software to perform.
> 
> @jason-p-pickering keep us honest, is this a good justification?

I see no good justification to not identify those rows in empty sheets as an error. I think whatever we do though, how we want the software to perform should be captured in a unit test. This is not the case currently for this particular issue, so I agree that for now, we should maybe just ignore it and deal with it more comprehensively in the next release.

jason-p-pickering (approved):

PR #703, opened 2023-06-15
DP-1071: COP23 OPUs Initial Setup

jacksonsj:
### Related Issues
- DP-1071
- DP-1088
- DP-1110

## Reviewer:
- [x] Clean linting.
- [x] Well commented.
- [ ] Build check passes.


R/packTool.R
```diff
@@ -49,12 +51,15 @@ packTool <- function(model_data = NULL,
   options("openxlsx.numFmt" = "#,##0")
 
   # Create data sidecar ####
-  d <- createDataPack(datapack_name = datapack_name,
-                                 country_uids = country_uids,
-                                 template_path = template_path,
-                                 cop_year = cop_year,
-                                 tool = tool,
-                                 d2_session = d2_session)
+  d <- suppressWarnings(
```
jason-p-pickering, line 54: Do we really want to suppress the warnings?

jacksonsj:
That column width issue seems to be coming from these two pieces of code:

https://github.com/pepfar-datim/datapackr/blob/DP-1071/R/packPSNUxIM.R#L653-L657
https://github.com/pepfar-datim/datapackr/blob/DP-1071/R/packPSNUxIM.R#L667-L670

If I comment these out, file writes just fine.

jacksonsj:
@jason-p-pickering - Ready for your review

R/packPSNUxIM.R
```diff
@@ -34,7 +34,11 @@ packPSNUxIM <- function(wb, # Workbook object
   rm(params, p)
 
   # if the cop year is not 2021 or 2022, stops and throws message. ####
-  stopifnot("Packing PSNU x IM tabs is not supported for the requested COP year." = cop_year %in% c(2021, 2022, 2023))
+  years.in.play <- c("PSNUxIM", "OPU Data Pack") %>%
```
jason-p-pickering, line 37: Please use snake case for variable names.

R/packPSNUxIM.R
```diff
@@ -651,10 +655,20 @@ packPSNUxIM <- function(wb, # Workbook object
                           heights = 0)
 
   ## Hide unused columns in left section ####
-  openxlsx::setColWidths(wb = r$wb,
-                         sheet = "PSNUxIM",
-                         cols = col.im.percents[1]:col.im.percents[2],
-                         hidden = FALSE)
+  # openxlsx::setColWidths(wb = r$wb,
```
jason-p-pickering, line 658: All of this seems very odd to me, but it looks like you found a solution. Can you remove the commented code? Not sure we need it if we found something which works.

jason-p-pickering (changes requested):
In general, I think it looks good. Have requested a few changes. I will put together a unit test in a separate branch.

R/packPSNUxIM.R
jacksonsj: Stashing here for future reference:
```
# openxlsx::setColWidths(wb = r$wb,
  #                        sheet = "PSNUxIM",
  #                        cols = col.im.percents[1]:col.im.percents[2],
  #                        widths = 10)
  # 6/27/23 SJJ - For some reason the above throws this error, while
  # removeColWidths does not:
  #
  # Error in sprintf("<col min=\"%s\" max=\"%s\" width=\"%s\" hidden=\"%s\" customWidth=\"1\"/>", :
  # arguments cannot be recycled to the same length
```

jacksonsj:
@jason-p-pickering - I think I've addressed everything. Mind taking another look?

R/createDataPack.R
```diff
@@ -9,6 +9,7 @@
 #'
 createDataPack <- function(datapack_name = NULL,
                            country_uids,
+                           season,
```
jason-p-pickering, line 12: No need to set a default argument value (NULL) here?

jason-p-pickering (approved):
Looks good to me. Just maybe check that one comment about the default arg

JordanBalesBAO:
So are we good to get this into dev? If so I will cut for the release.

R/createDataPack.R
jacksonsj: Because this function uses `check_params` at the top, and includes `season`, we don't need `NULL` default arg, because even an empty arg will get filtered and replaced with whatever the package default is, defined in `check_params`.

PR #723, opened 2023-08-02
Removes a filter that was blocking "dreams" data from making its way …

JordanBalesBAO:
…through the compare function

Removes a filter that was blocking "dreams" data from making its way through the compare function

### Related Issues
- DP-1129

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.

### DIFF
--- a/R/datapackVsDatim.R
+++ b/R/datapackVsDatim.R
@@ -70,10 +70,10 @@ compareData_DatapackVsDatim <-
            d2_session = dynGet("d2_default_session",
                                inherits = TRUE),
            datim_data = NULL,
-           datastreams = c("mer_targets", "subnat_targets", "impatt")) {
+           datastreams = c("mer_targets", "subnat_targets", "impatt", "dreams")) {
 
 
-
+    # HO5wSDKttn3
 # start main processing
 # start off with dedups included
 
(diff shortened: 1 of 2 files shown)


flopez-bao (approved):

R/datapackVsDatim.R
```diff
@@ -88,8 +88,7 @@ compareData_DatapackVsDatim <-
       dplyr::mutate(dataset = dplyr::case_when(dataset == "impatt" ~ "subnat_targets",
                                                dataset == "mer" ~ "mer_targets",
                                                dataset == "subnat" ~ "subnat_targets",
-                                               TRUE ~ dataset)) %>%
-      dplyr::filter(dataset %in% datastreams)
```
jason-p-pickering, line 92: @JordanBalesBAO I think I am OK with this, but maybe we are going too far? If you look at the function definition, one could also change the default parameters (or supply them) to include the `dreams` dataset, which I think would have the effect of including the DREAMS data. Maybe this is the better approach as opposed to removing this line of code? Otherwise, the `datastreams` parameter seems to have no effect. I am just not sure if there may be unintended side effects with pursuing this route?

jason-p-pickering (changes requested):

R/datapackVsDatim.R
JordanBalesBAO: I agree with your logic completely. So that was actually the first path I went down,  added 'dreams' to the datastream list then added on line 91 dataset == dreams ~ dreams. The wall I hit came down stream at 152-160 where getCOPDataFromDATIM() runs twice.  Any guidance, or do I just need to keep tinkering?

JordanBalesBAO:
Should be good to re-review.

jason-p-pickering (approved):

PR #794, opened 2024-02-09
create packForPAW function and tests

flopez-bao:
- create packForPAW function
- remove pop data pull from packForDATIM
- add relevant tests for both functions


flopez-bao:
Added a `createPAWExport` as a potential route option -- I think it may be the most straightforward after more testing, you can run the COP24 Cameroon datapack to see the differences in the final object, the PAW export should have the extra key pop indicators:

```
library(datapackr)
library(magrittr)

# Point to DATIM login secrets ####
secrets <- Sys.getenv("SECRETS_FOLDER") %>% paste0(., "datim.json")

datimutils::loginToDATIM(secrets)

output_folder <- Sys.getenv("OUTPUT_FOLDER") %>% paste0(., "COP24 Data Packs/")
model_data_path <- Sys.getenv("MODEL_DATA_PATH")
snuxim_model_data_path <- Sys.getenv("SNUXIM_MODEL_DATA_PATH")

# Unpack Submitted Data Pack ####
d <- datapackr::unPackToolSet(file.choose()) 


paw_ex <- createPAWExport(d)
datim_ex <- createDATIMExport(d)
```

JordanBalesBAO:
NOTE for myself, HAVE TO do in an ifelse in the app send to paw button based upon cop year.  This will go away after this year

R/createPAWExport.R
```diff
@@ -0,0 +1,61 @@
...
+      },
+      error = function(e) {
+        stop("An error occurred for type ", type, ": ", conditionMessage(e))
+      }
+    )
+  }
+
+
+  if (d$info$tool == "Data Pack") {
+    if (d$info$cop_year == 2022) {
```
JordanBalesBAO, line 27: Lets dump this 22 if statement.

JordanBalesBAO: Only once we confirm 2022 is no longer supported

JordanBalesBAO:
> NOTE for myself, HAVE TO do in an ifelse in the app send to paw button based upon cop year. This will go away after this year

Won't be necessary the way pack for paw was coded.

jason-p-pickering (changes requested):
I am not really totally comfortable to simply duplicate these two functions. They are essentially 95% the same, so why not just have a common function which handles the common parts, and then handle the specific parts in smaller bits of code?

PR #806, opened 2024-03-13
HOT FIX - getValidOrgUnits 2024

JordanBalesBAO:
### Related Issues
- DP-XXX
- etc.

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


JordanBalesBAO:
Searching for instances of datapackr::valid_OrgUnits instead of getValidOrgUnits(cop_year)

flopez-bao (commented):
Looks like for the generation script the `menu` function is acting a bit erratically when i try to run the script and it doesnt run the rest regardless of what i choose. Might be better to just have an ifelse statement with a `stop`, unless its just my computer:

Screenshot 2024-03-14 at 11 24 26 AM

flopez-bao:
I think use_cache needs to be `TRUE` to produce the rda files via the switch for cop_year

flopez-bao (approved):
This looks good now. I think long term we will want to revisit post cop for a longer term solution.

R/getPSNUs.R
```diff
@@ -27,7 +27,7 @@ getValidOrgUnits <- function(cop_year = NULL) {
          "2021" = valid_PSNUs,
          "2022" = valid_PSNUs,
          "2023" = valid_OrgUnits,
```
jason-p-pickering, line 29: Minor but maybe we should rename this to valid_OrgUnits_2023 for the sake of clarity? I think we can also deprecated 2021-2022? That will likely break tests, but we probably need to do this.

jason-p-pickering (approved):

PR #824, opened 2024-05-06
Release 7.4.4

JordanBalesBAO:
## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/unPackingChecks.R
```diff
@@ -935,6 +933,7 @@ checkInvalidPrioritizations <- function(sheets, d, quiet = TRUE) {
 #Extracts grey cells from Row3 for all sheets
 getCriticalColumns <- function()  {
 
+  #TODO: Should this be hardcoded like this with no paramaters to the function?
   template_file <- system.file("extdata/COP23_Data_Pack_Template.xlsx", package = "datapackr")
```
flopez-bao, line 937: can we make this into a ticket to explore a better way to do this? That way the TODO does not float around like previous todos. Judging from the look of it, this may be something we hardcode but recognize as needing probable updates down the road.

flopez-bao (approved):

R/unPackingChecks.R
jason-p-pickering: Yeah, I agree that this looks not good. I think the best place for this would be in the schema itself, rather than going back to the template and extracting out these "critical" columns each time?

jason-p-pickering (approved):

PR #842, opened 2024-05-31
added test for strip_wb_nas

flopez-bao:
- added test for strip_wb_nas
- some in line documentation for the function

### Related Issues
- DP-873

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


R/strip_wb_NAs.R
```diff
@@ -12,27 +12,28 @@
 #'
 strip_wb_NAs <- function(d) {
 
-  # d$tool$wb$sharedStrings[d$tool$wb$sharedStrings == "<si><t>NA</t></si>"] <- "<si><t>0</t></si>"
-
+  # find indices of sharedStrings matching "<si><t>NA</t></si>", adjusting for zero-based indexing
   NA_sharedString <- grep("<si><t>NA</t></si>", d$tool$wb$sharedStrings) - 1
```
jason-p-pickering, line 16: Is it maybe better to use grepl here since we are just looking for the presence of these strings?

R/strip_wb_NAs.R
```diff
@@ -12,27 +12,28 @@
...
+  # proceed only if such sharedStrings are found
   if (length(NA_sharedString) > 0) {
 
-   for (i in seq_along(d$tool$wb$worksheets)) {
-
-     might_be_NA <- d$tool$wb$worksheets[[i]]$sheet_data$v %in% NA_sharedString
+    for (i in seq_along(d$tool$wb$worksheets)) {
 
-     is_string <- d$tool$wb$worksheets[[i]]$sheet_data$t == 1
+      # identify cells with sharedStrings that might be "NA"
```
jason-p-pickering, line 23: I think it could be simpler to make this code inside a function and then write the unit test on that (outside of the loop)?

jason-p-pickering (changes requested):

flopez-bao:
@jason-p-pickering responding to your 2 comments - 

comment 1. seems to me the indices is what the function is looking for hence the grep. I can see in the [original code](https://github.com/pepfar-datim/datapackr/commit/9415fc2edffee9e0bd70cdec196cbb464810ae35) this was done as a response to the [647](https://github.com/pepfar-datim/Data-Pack/issues/647#issuecomment-587360871). Best not to change unless you have a specific solution in mind.

comment 2. i moved that portion into a function `process_sheet` declared inside `strip_wb_NAs`, not sure if that's what you meant, but does look cleaner and did lapply instead of for loop.

jason-p-pickering:
Yes I see now what you mean on closer inspection of the code. Looks good to me now.

JordanBalesBAO (approved):

JordanBalesBAO (approved):

PR #863, opened 2024-08-13
feat: Add tool as an optional argument in message queue

jason-p-pickering:
### Related Issues
- DP-1321

- When merging datapacks, we may have tests or messages which come from one tool or the other. When we merge the message queues, it was not always clear which tool the message was applicable to. When merging datapacks, we will include the tool as a column in tests and messages to help to better identify the origin. 
-  This will not really help when merging two data packs. We may need to think a bit more about this use case, but in general, for end users, they will be looking at a DataPack and a PSNUxIM tool, so it should be fairly straightforward for them to identify which tool the message/test is applicable to. 

## Reviewer:
- [ ] Clean linting.
- [ ] Well commented.
- [ ] Build check passes.


jason-p-pickering:
@flopez-bao @JordanBalesBAO this is ready for review now.

flopez-bao (approved):

